PR #1, opened 2020-05-19
Decouple MANA code from coordinator code

rohgarg:
These are the first steps to decouple the coordinator and the MANA code.

I have moved all of the MANA-specific code from `dmtcp_coordinator.cpp` to a separate set of files that are compiled along with the MANA plugin code. This code can be linked independently with the `dmtcp_coordinator` binary. I have added appropriate dependencies in the coordinator `Makefile`. It seems like a cleaner design.

The build process for the coordinator is still not fully automated yet, but it seems better than having a separate compilation script (`compile-coord-cori.sh`) just for the coordinator. After applying these patches, the steps for compiling the coordinator would be as follows:

```bash
$ cd $DMTCP_ROOT
$ ./configure --enable-debug
$ cd contrib/mpi-proxy-split/ && make -j # First, build the plugin
$ cd $DMTCP_ROOT && make -j # Next, finish building the rest of DMTCP
```

Note that the plugin can be compiled independent of DMTCP and must be built first. This creates the relevant file in the plugin directory (`mana_coordinator.o`) required to build the dmtcp_coordinator binary

Since I cannot directly compile and test on Cori, I am not directly pushing this change to the refactoring branch and instead created this PR. Could someone please test this and see if it works?

### DIFF
--- a/compile-coord-cori.sh
+++ b/compile-coord-cori.sh
@@ -1,8 +1,0 @@
-#!/bin/bash
-cd src/
-CC -DHAVE_CONFIG_H -I. -I../include  -I../include -I../jalib -Wall -DDEBUG -fPIC \
-   -std=c++11 -g3 -O0 -DDEBUG -MT dmtcp_coordinator.o -MD -MP -MF .deps/dmtcp_coordinator.Tpo \
-   -c -o dmtcp_coordinator.o dmtcp_coordinator.cpp
-# Next line no longer needed.  'make -j' does it.
-# g++ -g3 -O0 -o ../bin/dmtcp_coordinator dmtcp_coordinator.o lookup_service.o restartscript.o \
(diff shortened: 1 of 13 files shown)


rohgarg:
@gc00 If we try to build DMTCP first, the coordinator will fail to build, because it needs this object file (`mana_coordinator.o`) from the plug-in directory.

I already added the dependency for building the `dmtcp_coordinator` binary on `mana_coordinator.o`. The problem is that the top level Makefile doesn't yet know how to cd into the plug-in directory and compile the plug-in source files using the MPI compiler.

gc00:
Hi Rohan,
    I didn't have a chance to install libxml2.a in my setup.  Probably you already have it.
Does MANA work locally for you now?

Also, I propose that we use configure-mana to add new configurations for other MPIs, too.
We can comment out all but the one that we're using.  It's not a clean way to do it, but it's good enough for this early development stage.

Best,
- Gen

________________________________________

PR #2, opened 2020-05-19
Refactor: Create autoconf variables for MPI, to run with other MPI implementations

gc00:
Hi @rohgarg,

This is a first attempt at making MANA portable across different MPI implementations.

Look at `DMTCP_ROOT/configure-mana` for an example of how to use it.  After configuring, it will then use the Makefile variables found in `DMTCP_ROOT/contrib/mpi-proxy-split/Makefile_config`.

I tested this with MPICH.  The `configure-mana` file is set for my MPICH build on login (CentOS).

Mostly, I can do 'make' in `contrib/mpi-proxy-split`.  But in lower-half, 'make' fails to compile the `lh_proxy` executable, with the error message:
`/usr/bin/ld: cannot find -lxml2`
In my next debugging session, I'll look for why it needs libxml2.a.  Apparently, `mpicc` translates `-lmpi -lrt` into `-lmpi -lm -lxml2 -lpthread -lrt`.  I'm not sure why.  Looking at `sh -x mpicc ...`, I don't see an `-lxml2` when compiling simple MPI programs in the test subdirectory.

**IN THIS UPDATED PR, THE BUG BELOW HAS BEEN FIXED NOW BY THE SECOND COMMIT:**
_What you'll see is that `contrib/mpi-proxy-split/mpi-wrappers/mpi_dummy.c` fails to compile, because the mpi.h supplied by MPICH is incompatible with our assumed mpi.h in that file.  Maybe you can diagnose this further, @rohgarg.  Feel free to extend this branch and merge it, if you can figure out `mpi_dummy.c`.  Developing on a local machine will be faster than developing on Cori, where 'srun' can take a long time to start up._

Also, as a small side note, I saw that `test/autotest_config.py` was placed under git version control.  This seems to be a mistake, since we already have `test/autotest_config.py.in`.  But you know your intentions better than me.  Could you also take a look at that?


rohgarg:
> The configure-mana file is set for my MPICH build on login (CentOS). What you'll see is that contrib/mpi-proxy-split/mpi-wrappers/mpi_dummy.c fails to compile, because the mpi.h supplied by MPICH is incompatible with our assumed mpi.h in that file.

I'm aware of the issue, @gc00. The problem occurs because some of the MPI functions have a slightly different prototype in the two implementations, which is strange. :-) I have had to manually edit the wrapper prototypes in the plugin to make it work so far. Anyway, I will try to create a more general fix to make it work.

rohgarg:
@gc00: I checked my build of mpich-3.3.2 on a CentOS7 machine. I don't have libxml2 installed. I configured and built mpich with the following options
```bash
$ ./configure --enable-g=dbg --prefix=/home/rohgarg/sws
```

Here's what the `config.log` says:

```
configure:38772: checking for LIBXML2
configure:38780: $PKG_CONFIG --exists --silence-errors "libxml-2.0"
configure:38783: $? = 1
configure:38799: $PKG_CONFIG --exists --silence-errors "libxml-2.0"
configure:38802: $? = 1
No package 'libxml-2.0' found
12265 configure:38832: result: no
```

Th next code snippet is from the header file (`src/include/mpichconf.h`) it generates.

```C
/* Define to 1 if you have the `libxml2' library. */
/* #undef HWLOC_HAVE_LIBXML2 */
```

I was looking at the options used by `mpicc` while trying to link the lower half binary. Here's the full command line:

```bash
/usr/libexec/gcc/x86_64-redhat-linux/4.8.5/collect2 --build-id --no-add-needed --hash-style=gnu -m elf_x86_64 -static -o lh_proxy /usr/lib/gcc/x86_64-redhat-linux/4.8.5/../../../../lib64/crt1.o /usr/lib/gcc/x86_64-redhat-linux/4.8.5/../../../../lib64/crti.o /usr/lib/gcc/x86_64-redhat-linux/4.8.5/crtbeginT.o -L/home/rohgarg/sws/lib -L/usr/lib/gcc/x86_64-redhat-linux/4.8.5 -L/usr/lib/gcc/x86_64-redhat-linux/4.8.5/../../../../lib64 -L/lib/../lib64 -L/usr/lib/../lib64 -L/usr/lib/gcc/x86_64-redhat-linux/4.8.5/../../.. -Ttext-segment 0xE000000 --wrap __munmap --wrap shmat --wrap shmget lh_proxy.o libproxy.a -lmpi -lrt -lpthread -lc libproxy.a -rpath /home/rohgarg/sws/lib --enable-new-dtags -lmpi -lm -lpthread -lrt --start-group -lgcc -lgcc_eh -lc --end-group /usr/lib/gcc/x86_64-redhat-linux/4.8.5/crtend.o /usr/lib/gcc/x86_64-redhat-linux/4.8.5/../../../../lib64/crtn.o```

**Comment by rohgarg** [comment]
@gc00: I looked at mpich's source more carefully. It's the `hwloc` library (a dependency of mpich, and shipped along with the mpich source) that requires libxml2. However, `hwloc` does detect if libxml2 is not available on the system and can be built with libxml2 support.

On my CentOS7 system, with no libxml2 installed, it indeed builds without libxml2.

```bash
/home/rohgarg/sws/mpich-3.3.2/src/hwloc/hwloc $ ls topology-xml*
topology-xml.c  topology-xml.lo  topology-xml.o  topology-xml-libxml.c  topology-xml-nolibxml.c  topology-xml-nolibxml.lo  topology-xml-nolibxml.o
```

Note that only the `*-nolibxml` versions of the source files are compiled. The `*-libxml` versions are not compiled, since the `configure` script did not find any libxml2 on the system (see my [comment above](https://github.com/mpickpt/mana/pull/2#issuecomment-631644111)).

gc00:
@rohgarg, I think the problem with MPICH is that hwloc tests to see if libxml2.so is available.  It's like you say, but it looks specifically for the .so version.  If so, then MPICH is built with a dependency on libxml2.  But if libxml2.a is not available, then MPICH will fail to support 'mpicc -static'.  Since dekaksi had libxml2.so, but not libxml2.a, I tried installing the package libxml2-dev.  After that, 'mpicc -static' did not complain about libxml2.a anymore.  So, that's good.

But then the next problem is that mpicc (when linking) still complained:

```
/usr/bin/ld: cannot find -lcr
```

Sure enough, dekaksi has libcr.so present, but not libcr.a  It did have libcr.la and libcr_omit.la and libcr_omit.so.  Apparently, 'mpicc -static' is not recognizing the presence of libcr.la.  Possibly, one could link with libcr.la by using libtool, but that gets really ugly.  Maybe MANA should just provide its own libcr_omit.a file and link with that, to fix this issue.  But libcr.so has a lot of dynamic symbols.  It also seems ugly for us to have to add those.

In the end, I found this solution to use the standard Ubuntu package for mpich on dekaksi:
```
[ WILL FILL IN soon; I need to look it up. ]
```

What happens on your CentOS7 machine?  Do you have libcr.so present, and do you have libcr.a present?

rohgarg:
> What happens on your CentOS7 machine? Do you have libcr.so present, and do you have libcr.a present?

I had configured and built mpich from source. I don't seem to have a libcr.so or libcr.a anywhere on my system.

rohgarg:
In my case, during configure and build time, it detects that libcr is not present and builds accordingly.

From `src/include/mpichconf.h`:
```C
/* Define to 1 if you have the `cr' library (-lcr). */
/* #undef HAVE_LIBCR */
```

From `src/pm/hydra/config.log`:
```
configure:25444: checking checkpointing library
configure:25450: result: blcr
configure:25575: checking libcr.h usability
configure:25575: gcc -std=gnu99 -c    -g -O2    -I/home/rohgarg/sws/mpich-3.3.2/src/mpl/include -I/home/rohgarg/sws/mpich-3.3.2/src/mpl/include -I/home/rohgarg/sws/mpich-3.3.2/src/openpa/src -I/home/rohgarg/sws/mpich-3.3.2/src/openpa/src -D_REENTRANT -I/home/rohgarg/sws/mpich-3.3.2/src/mpi/romio/include conftest.c >&5
conftest.c:226:19: fatal error: libcr.h: No such file or directory
 #include <libcr.h>
...
configure:25575: result: no
configure:25575: checking libcr.h presence
```

From `src/pm/hydra/include/hydra_config.h`:
```C
/* Define if blcr is available */
/* #undef HAVE_BLCR */
```
I'm guessing since you had directly installed the mpich package through the distro's package manager, it installed `libcr.so` as a dependency. I was going through BLCR's source, and they doesn't compile for static linking by default. That's why they don't ship `libcr.a` by default.

gc00:
@rohgarg,
    Our messages crossed, as I tried to edit my own message for version-2.
    So, I think we've solved the enigma.  Shall we use:
```
ln -s libproxy.a libcr.a
mpicc -static -Wl,--warn-unresolved-symbols -Wl,-Ttext-segment -Wl,0xE000000 -Wl,--wrap,__munmap -Wl,--wrap,shmat -Wl,--wrap,shmget -o lh_proxy lh_proxy.o libproxy.a -L. -lcr > /dev/null
```

We can add a comment about what the issue is.  And finally, we should auto-detect if the user has libxml2.so but not libxml2.a, and then ask the user to install libxml2-dev.

If we do this, we don't have to ask the user to compile their own MPICH.

PR #3, opened 2020-05-21
mana-plugin: Don't use mpicc/mpicxx for linking upper half libs/execs

rohgarg:


rohgarg:
@ItsPrashant, @gc00, @JainTwinkle: I realized that there was a minor bug in the way we were linking the upper half libraries (the plugin library and libdummympi.so) that would create a dependency on the real MPI library. This would cause the runtime linker/loader to load the real MPI library into the upper half.

Please test this change on Cori before merging it.

PR #5, opened 2020-05-26
Refactor: mtcp_restart does set_lh_mem_range, instead of lower-half doing it

gc00:
**_[ This works now.  Please review. ]_**

setLhMemRange() has been moved to src/mtcp/mtcp_split_process.c.  (And another version of it is defined in contrib/mpi-proxy-split/split-process.cpp.)  Please review this, so I can push it in, and build on top of it.

Once this is in 'refactoring', I can then test PR #2 and ask for that also to be reviewed.  With that in place, I will then modify setLhMemRange() to set the lh_memory_range based on other memory layouts, including CentOS/Ubuntu.

Right now, MANA is not working on CentOS/Ubuntu because the lh_mem_range is conflicting with the stack on CentOS/Ubuntu.  The libs and stack, there, have only 128 MB free between them.

We can then port to other layouts of the stack, vdso, libs, etc..  So, the choice of where to put the LhMemRange() will be based on an analysis of the current memory layout.

This code continues to work on Cori., and should not break anything.

Other simplifications were also made in the code:
  1. lh_proxy reads the lh_mem_range on its stdin, and writes back the lh_info struct on its stdout.  So, lh_proxy (the child process) doesn't need special coding with pipes; that's only for the parent process now.
  2.  This also means that we don't need to pass buf as arg[1] of lh_proxy, holding the pipe fd that will be used by lh_proxy.
 3.  We now have only one routine, setLhMemRange() (in mtcp_restart), instead of two of them in the contrib/mpi-proxy-split plugin.
 4. Various comments were added to make clear the logic flow in splitProcess(), and how the lh_info struct and lh_mem_range are passed back and forth.
 5. 'info' and 'pdlsym' are now declared and allocated in lower-half-api.h and split-process.cpp.  So, now split-process.cpp no longer depends on mpi_plugin.h.

### DIFF
--- a/src/mtcp/mtcp_split_process.c
+++ b/src/mtcp/mtcp_split_process.c
@@ -163,8 +163,10 @@ startProxy(char *argv0, char **envp)
 
     case 0: // in child
     {
-      mtcp_sys_close(pipefd_out[0]); // close reading end of pipe
-      // FIXME:  This could be done more simply by writing to stdin of lh_proxy.
+      mtcp_sys_dup2(pipefd_out[1], 1); // Will write lh_info to stdout.
+      mtcp_sys_close(pipefd_out[1]);
+      mtcp_sys_close(pipefd_out[0]); // Close reading end of pipe.
+      // FIXME:  Remove buf after verifying can write to stdout instead.
       char buf[10];
       mtcp_itoa(pipefd_out[1], buf);
       char* args[] = {"NO_SUCH_EXECUTABLE", buf, NULL};
--- a/src/mtcp/mtcp_split_process.c
+++ b/src/mtcp/mtcp_split_process.c
@@ -22,12 +22,19 @@
 
 int mtcp_sys_errno;
 LowerHalfInfo_t info;
+// FIXME:  The value of pdlsym must be passed from mtcp_restart to
+//         the upper half, so that NEXT_FNC() in mpi-wrappers in libmana.so in 
+//         can use the updated pdlsym in case the address of the lower half
+//         in the restarted process has changed.
+//         But the lower half is loaded at a fixed address.  So, the address of
+//         pdlsym can be a fixed address that is unused by the upper half.
+//         In this case, we don't need to pass pdlsym to the upper half.
+static proxyDlsym_t pdlsym; // initialized to (proxyDlsym_t)info.lh_dlsym
 MemRange_t *g_lh_mem_range = NULL;
 static MemRange_t g_lh_mem_range_buf;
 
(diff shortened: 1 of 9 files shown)


rohgarg:
> lh_proxy reads the lh_mem_range on its stdin, and writes back the lh_info struct on its stdout. So, lh_proxy (the child process) doesn't need special coding with pipes; that's only for the parent process now.

I don't understand the motivation for doing this. In particular, I don't understand why we need to exchange lh_mem_range information between the main process and the transient proxy process. The memory range info is relevant only *after* we've copied the lower half bits into the main process and we're initializing and calling functions in the lower half from the upper half.

We can simply define a new API in the lower half to set the memory range and call that function from the upper half, similarly to other APIs in lower_half_api.h.

Or is there a case where the lower half needs this information even while it's part of the transient proxy process?

> We now have only one routine, setLhMemRange() (in mtcp_restart), instead of two of them in the contrib/mpi-proxy-split plugin.

I didn't understand this point. We still have the duplicate definitions in `contrib/mpi-proxy-split/split_process.cpp` and `src/mtcp/mtcp_split_process.c`.

src/mtcp/mtcp_split_process.c
```diff
@@ -1,3 +1,4 @@
+// FIXME:  Why do we have all these libc headers if we're not using libc here?
```
rohgarg, line 1: You are right, we don't need all the libc headers here. We only need a few of them in order to compile this file.

Here's a minimal list of header includes we do need in order to compile this file; every other include can be removed.

```C
#include <asm/prctl.h> // For ARCH_GET_FS, ARCH_SET_FS
#include <sys/prctl.h> // For ARCH_GET_FS, ARCH_SET_FS
#include <sys/personality.h> // For ADDR_NO_RANDOMIZE
#include <ucontext.h> // For ucontext_t
#include <sys/uio.h> // For struct iovec
#include <link.h> // For ElfW, auxv_t
```

gc00:
@rohgarg , because I'm basing my development on top of this PR, it was getting complicated.  So, I'm pushing this in now.  I looked back, and I think I placed info and pdlsym and setLhMemRange() in split-process.cpp in the plugin because their analogs were in mtcp_split_process.c.  I agree that these should probably be in mptc_restart.c and mpi_plugin.cpp.  I can do that later as part of a cleaning-up exercise.  For now, I just want the same architecture for the two cases.

By the way, I think I can now make MANA work with MPICH on dekaksi using the test-dekaksi branch.  I'll test that further before setting up a PR for it.

PR #6, opened 2020-05-27
mana-plugin: Use C++ atomics for recording sends/recvs

rohgarg:


rohgarg:
It's just using atomic operations to increment/decrement the counters. It doesn't use any locks. Rather than using the non-portable GCC built-ins for CAS/atomics, C++ provides these portable abstractions. My intent was to remove the locks for efficiency and reducing code size.

I'm not strongly inclined either way. We can keep this PR open and decide later if it's required.

gc00:
Let's talk about it directly.  It's a tension between keeping the core knowledge needed by a first-year PhD student to a minimum,;and writing clean, efficient, maintainable code.

gc00:
@rohgarg, the C++ atomics seem to be using mutexes way down lower in the software stack.  We've found that mutexes are expensive, since they generate a kernel call.  When frequent MPI calls are made (several thousand per second), the overhead ofa  kernel call is at least a few percent.

We are now using rwlock as cheaper.  The rdlock does not usually generate a kernel call.

You're welcome to update this PR -- or else we will continue to separately review the MANA code and add `pthread_rwlock_rdlock` strategically, in the user thread; while using `pthread_rwlock_wrlock` in the ckpt thread.  This is valid as long as only one user thread (and no other user thread) is making the relevant MPI call.

PR #8, opened 2020-05-29
Testing mana with hugepages

ItsPrashant:
This branch is still under development. I need to test these changes further, but it seems to work for now.

### DIFF
--- a/src/writeckpt.cpp
+++ b/src/writeckpt.cpp
@@ -268,7 +268,12 @@ mtcp_writememoryareas(int fd)
        * Posix and SysV shared memory segments can be mapped as /XYZ
        */
     }
-
+    /* Check if the area is a huge page
+     * If yes, add MAP_HUGETLB to the flags
+     */
+    if(strstr(area.name, "anon_hugepage")){
+      area.flags|=MAP_HUGETLB;
+    }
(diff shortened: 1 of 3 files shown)


src/writeckpt.cpp
```diff
@@ -268,7 +268,12 @@ mtcp_writememoryareas(int fd)
        * Posix and SysV shared memory segments can be mapped as /XYZ
        */
     }
-
+    /* Check if the area is a huge page
+     * If yes, add MAP_HUGETLB to the flags
+     */
+    if(strstr(area.name, "anon_hugepage")){
```
rohgarg, line 274: A huge page won't always be "anon_hugepage". See the three example applications from the kernel source.
 - [map_hugetlb.c](https://elixir.bootlin.com/linux/latest/source/tools/testing/selftests/vm/map_hugetlb.c): Private anonymous mmap
 - [hugepage-mmap.c](https://elixir.bootlin.com/linux/latest/source/tools/testing/selftests/vm/hugepage-mmap.c): Mapping in hugetlbfs
 - [hugepage-shm.c](https://elixir.bootlin.com/linux/latest/source/tools/testing/selftests/vm/hugepage-shm.c): Mapping in SysV shared pages with

PR #12, opened 2021-03-23
Updated drain_send_recv_packets.cpp to properly use the mutexes

NeilResnik:
This commit makes the various calls in drain_send_recv_packets.cpp properly take the mutexes associated with certain data structures, and properly prevent issues with modifying the vector that is being iterated over by calling certain wrapper functions within the iteration. Previously, due to the fact the program shouldn't be threaded at this point, the mutexes were ignored, which is not only inconsistent, but would hide a future bug if multiple threads were running. Furthermore, the functions made calls to wrappers which modified the list they were iterating over, creating bugs in that regard as well. This commit fixes that so that the underlying code modifying the vector being iterated over no longer results in bugs (through the use of the erase method provided by std::vector), and properly takes the mutexes associated with the data structures before using them.


gc00:
This PR seems to be obsolete.  I'm closing this, as irrelevant to the current MANA.

PR #13, opened 2021-03-23
Fix issues with the restore*Free Calls differing from the wrappers.

NeilResnik:
This fixes a bug where when a checkpoint is restored from, the semantics of the various MPI structure calls does not match the wrapper functions. For MPI_Comm_free, MPI_Group_free, MPI_Type_free, and MPI_Op_free, when the structure is freed, the record of it is not deleted. This is explained by comments in the appropriate wrapper files, for example in mpi-wrappers/mpi_comm_wrappers: "NOTE: We cannot remove the old comm from the map, since we'll need to replay this call to reconstruct any other comms that might have been created using this comm.". This semantic was not maintained in the restoreCommFree (and similar) functions, which did remove the structure from the map. This meant that when a structure was freed, if you created a checkpoint, loaded from that checkpoint, and then created a checkpoint from the loaded checkpoint, that the second created checkpoint (the one created from the original loaded checkpoint) is corrupt. That is, since the freed structure was removed from the map, it couldn't properly recreate the MPI state. This often manifested as it attempting to free a non-existent structure, resulting in internal MPI errors. This fixes this so the restore calls match the semantics of the original wrapper calls, preventing this bug.


gc00:
This commit has now been duplicated by 13e40285a763738d5d30bfe4d49b898566311  [`Avoid removing virtual ids (Group, ...) at restart`].  That commit was pushed into refactoring by @xuyao0127 .  It's very unfortunate that we didn't see this PR in time.  It fixes that same bug as @xuyao0127 .  We need better communication.

Anyway, if @ItsPrashant agrees, then let's close this PR without committing it.

gc00:
On April 20, 2021, I already asked the authors to close this PR, as a duplicate of a different one.

I'm closing the PR now.

PR #17, opened 2021-07-27
Replace minimal TLS/TCB header

jungan:


contrib/mpi-proxy-split/split_process.h
```diff
@@ -66,7 +66,10 @@ static int fsaddr_initialized = 0;
 // #define BUF_SIZE 256
 // #define BUF_SIZE 128
 #define BUF_SIZE 64
-static char *fsaddr_buf[BUF_SIZE + 3 * sizeof(void *)];
+// used part of tcb header:
+// (gdb) p (char *) &((struct pthread *) $rsp)->header.__unused2 - $rsp
+static const size_t TCB_HEADER_SIZE = 120;
```
rohgarg, line 71: How did we come up with this magic number (120)? Is it possible to make it a bit more dynamic? If not, please update the comment above, explaining what this represents. I understand that the intention here is to copy the relevant bits from the TCB.

contrib/mpi-proxy-split/split_process.h
```diff
@@ -76,12 +79,36 @@ static inline void SET_LOWER_HALF_FS_CONTEXT() {
...
+  // at the old location
+
+  // changing the location breaks things that keep pointers into the struct,
+  // like linked lists. TCB has a lot of internal self-pointers and lists, so
+  // it is advantageous to keep it in place. TLS usually has less (application
+  // dependent), and must be moved because it can be accessed relative to fs
+  // with no indirection. So we change self but not header.
+
+  // change self pointer to new application half TLS location
+  ((void **)(uh_fsaddr + BUF_SIZE))[0] = (void *) (uh_fsaddr + BUF_SIZE);
```
rohgarg, line 103: Don't we need to update the other two pointers -- dtv and header?

rohgarg (changes requested):
Thanks for the patch, @jungan. I have some minor comments/questions below.

contrib/mpi-proxy-split/split_process.h
jungan: The original author got this number from gdb:
```
(gdb) p (char *) &((struct pthread *) $rsp)->header.__unused2 - $rsp
$1 = 120

{tcb = 0x403e5f <rpl_nanosleep+175>, dtv = 0x14, self = 0x0, multiple_threads = -7476, gscope_flag = 32767, sysinfo = 13871592825318420480, stack_guard = 140737353938608, pointer_guard = 140737488346608, vgetcpu_cache = {140737488346720, 4209856}, __unused1 = 20, __glibc_unused1 = 0, __private_tm = {0x0, 0x7fffffffdf90, 0xc081cc2e05f63000, 0x7fffffffde60, 0x7fffffffdf98}, __unused2 = 4201008, __glibc_unused2 = {{{i = {4200141, 0, 4212990, 0}}, {i = {4212978, 0, 0, 0}}, {i = {0, 1077149696, 0, 0}}, {i = {0, 1077149696, -7474, 32767}}}, {{i = {100020224, -1065235410, 0, 0}}, {i = {0, 0, 0, 0}}, {i = {4200368, 0, -8320, 32767}}, {i = {0, 0, 0, 0}}}, {{i = {-140315307, 32767, 0, 0}}, {i = {-8312, 32767, 0, 2}}, {i = {4199696, 0, 0, 0}}, {i = {-653531328, 1832543198, 4200368, 0}}}, {{i = {-8320, 32767, 0, 0}}, {i = {0, 0, 1684793152, -1832543199}}, {i = {856811328, -1832547173, 0, 0}}, {i = {0, 0, 0, 0}}}, {{i = { 0, 0, 0, 0}}, {i = {0, 0, 0, 0}}, {i = {0, 0, 4200368, 0}}, {i = {-8320, 32767, 0, 0}}}, {{i = {4200409, 0, -8328, 32767}}, {i = {28, 0, 2, 0}}, {i = {-7491, 32767, -7476, 32767}}, {i = {0, 0, -7473, 32767}}}, {{i = {-7462, 32767, -7445, 32767}}, {i = {-7425, 32767, -7394, 32767}}, {i = {-7361, 32767, -7341, 32767}}, {i = {-7325, 32767, -7311, 32767}}}, {{i = {-7283, 32767, -7256, 32767}}, {i = {-7238, 32767, -7214, 32767}}, {i = {-7182, 32767, -7132, 32767}}, {i = {-7104, 32767, -7081, 32767}}}}, __padding = {0x7fffffffe462, 0x7fffffffeb1a, 0x7fffffffeb4a, 0x7fffffffeb59, 0x7fffffffeb65, 0x7fffffffec4e, 0x7fffffffec69, 0x7fffffffec7c}}
```
While I also checked the source code in glibc:
```
typedef struct
{
  void *tcb;            /* Pointer to the TCB.  Not necessarily the
                           thread descriptor used by libpthread.  */
  dtv_t *dtv;
  void *self;           /* Pointer to the thread descriptor.  */
  int multiple_threads;
  int gscope_flag;
  uintptr_t sysinfo;
  uintptr_t stack_guard;
  uintptr_t pointer_guard;
  unsigned long int vgetcpu_cache[2];
  /* Bit 0: X86_FEATURE_1_IBT.
     Bit 1: X86_FEATURE_1_SHSTK.
   */
  unsigned int feature_1;
  int __glibc_unused1;
  /* Reservation of some values for the TM ABI.  */
  void *__private_tm[4];
  /* GCC split stack support.  */
  void *__private_ss;
  /* The lowest address of shadow stack,  */
  unsigned long long int ssp_base;
  /* Must be kept even if it is no longer used by glibc since programs,
     like AddressSanitizer, depend on the size of tcbhead_t.  */
  __128bits __glibc_unused2[8][4] __attribute__ ((aligned (32)));

  void *__padding[8];
} tcbhead_t;

# ifdef __ILP32__
/* morestack.S in libgcc uses offset 0x40 to access __private_ss,   */
_Static_assert (offsetof (tcbhead_t, __private_ss) == 0x40,
                "offset of __private_ss != 0x40");
/* NB: ssp_base used to be "long int __glibc_reserved2", which was
   changed from 32 bits to 64 bits.  Make sure that the offset of the
   next field, __glibc_unused2, is unchanged.  */
_Static_assert (offsetof (tcbhead_t, __glibc_unused2) == 0x60,
                "offset of __glibc_unused2 != 0x60");
# else
/* morestack.S in libgcc uses offset 0x70 to access __private_ss,   */
_Static_assert (offsetof (tcbhead_t, __private_ss) == 0x70,
                "offset of __private_ss != 0x70");
_Static_assert (offsetof (tcbhead_t, __glibc_unused2) == 0x80,
                "offset of __glibc_unused2 != 0x80");
# endif
```
I think we stopped at
```
  /* The lowest address of shadow stack,  */
  unsigned long long int ssp_base;
```
From testing, it looks fine, but I don't have enough understanding if this is all we need to do, or we actually can do less.

contrib/mpi-proxy-split/split_process.h
jungan: I think we don't have to
For dtv, because mana statically linked libs in the lower half, so this won't change, if mana will use dynamical linking in the future, link it with `-ftls-model=initial-exec` can also avoid the problem.
For `self`, we can let it access the TCB in the old place, we may don't need to copy back TCB when restore, let me make the change and give it another try.

PR #24, opened 2021-10-26
[MANA-36] mana: update unit tests to use Google test framework

rohgarg:
Currently, only the group test works; all other tests are failing.
The drain test doesn't even compile since the drain code has changed
significantly since the last time I looked at it.

### DIFF
--- a/contrib/mpi-proxy-split/unit-test/record-replay-comm-test.cpp
+++ b/contrib/mpi-proxy-split/unit-test/record-replay-comm-test.cpp
@@ -1,6 +1,4 @@
-#include <cppunit/TestFixture.h>
-#include <cppunit/extensions/HelperMacros.h>
-#include <cppunit/ui/text/TestRunner.h>
+#include <gtest/gtest.h>
 
 #include <mpi.h>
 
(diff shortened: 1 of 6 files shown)


gc00:
@jungan , Could you take a look at this PR?  Thanks.  Your experience with the Google test framework is ideal for reviewing this.

We want to enable people to start writing more unit tests (e.g., Illio for the startup phase of MANA, with different memory layouts on CentOS and Cori).

contrib/mpi-proxy-split/unit-test/record-replay-comm-test.cpp
```diff
@@ -31,87 +28,80 @@ class CommTests : public CppUnit::TestFixture
...
+}
 
 int
 main(int argc, char **argv)
 {
-  CppUnit::TextUi::TestRunner runner;
-  runner.addTest(CommTests::suite());
-  return runner.run("", false, true, false) ? 0 : -1;
+  ::testing::InitGoogleTest(&argc, argv);
+  return RUN_ALL_TESTS();
```
jungan, line 106: NIT: Do you think it's better to just use one `main` which includes all the test suites?
Or one test binary for each independent module? Like one for MPI functions, one for LH utilities, one for dmtcp, etc.

jungan (approved):
Thanks a lot for building this nice framework for the unit test!
Although I'm not familiar with some MPI tests, the overall structure is clear to me, only one open question that if we want to have fewer test binaries.
I will also try it out by myself once it's merged.

contrib/mpi-proxy-split/unit-test/record-replay-comm-test.cpp
rohgarg: It's a little tricky because testing different functionality requires different set up. But we can see how to do that later.

PR #26, opened 2021-11-01
[MANA-48] Dynamically find a memory region for LH to avoid overlap

fyshhh:
This pull request fixes a bug where the assigned lower half memory region would overlap with some of the libraries before the stack on CentOS, as it is automatically assigned one gigabyte before the stack on Cori. It now searches for the first available two gigabyte memory region to assign the memory region to - we use two gigabytes to give it some buffer space; it also leaves a one gigabyte buffer region if it's placed after the heap or before the stack in case either of them grow. From what I've seen, this ends up being much before the hardcoded location, usually just after the heap.

This has been tested on both Cori and CentOS by running all the programs in `contrib/mpi-proxy-split/test` directories, and the results seen are the same as without this pull request. If an available region is not found, it fails an assertion, so this should mean that the algorithm is working, but to be safe I have inspected the memory map of these programs and found them to be fine.

### DIFF
--- a/contrib/mpi-proxy-split/split_process.cpp
+++ b/contrib/mpi-proxy-split/split_process.cpp
@@ -47,10 +47,7 @@
 #include "procmapsutils.h"
 #include "util.h"
 #include "dmtcp.h"
-
-#ifndef CENTOS
-# define CENTOS
-#endif
+#include "mtcp/mtcp_util.h"
 
 static unsigned long origPhnum;
 static unsigned long origPhdr;
@@ -348,64 +345,63 @@ startProxy()
 }
 
 // Rounds the given address up to the nearest 2MB.
-static inline unsigned long long
-alignMemAddr(unsigned long long addr)
+static inline uint64_t alignMemAddr(uintptr_t *addr, uintptr_t *size)
 {
-  return (addr + 0x1fffff) & ~0x1fffff;
+  return (*addr + 0x1fffff) & ~0x1fffff;
 }
 
-// This sets the address range of the lower half by searching through the memory
-// region for the first available free region.
+// This sets the address range of the lower half dynamically by searching
+// through the memory region for the first available free region.
 static void
-dynamicMemRangeHelper(MemRange_t *lh_mem_range, const uint64_t *region_size)
+findLHMemRange(MemRange_t *lh_mem_range)
 {
-  const uint64_t ONE_GB = 0x40000000;
-  
   bool is_set = false;
 
-  char perms[8];
-  unsigned long offset;
-  char device[8];
-  unsigned long inode;
+	Area area;
   char prev_path_name[PATH_MAX];
   char next_path_name[PATH_MAX];
+	uint64_t prev_addr_end;
+  uint64_t next_addr_start;
+  uint64_t next_addr_end;
 
-  unsigned long long prev_addr_start;
-  unsigned long long prev_addr_end;
-  unsigned long long next_addr_start;
-  unsigned long long next_addr_end;
+  int mapsfd = open("/proc/self/maps", O_RDONLY);
+	JASSERT(mapsfd >= 0)(JASSERT_ERRNO).Text("Failed to open proc maps");
 
-  // FIXME: fopen and fscanf may call malloc, which might call mmap.
-  // For now, we hope it doesn't change the calculation.
-  FILE* f = fopen("/proc/self/maps", "r");
-  if (f) {
-    fscanf(f, "%llx-%llx %s %lx %s %lx %s\n",
-           &prev_addr_start, &prev_addr_end, perms, &offset, device, &inode, prev_path_name);
-    while (fscanf(f, "%llx-%llx %s %lx %s %lx %s\n",
-           &next_addr_start, &next_addr_end, perms, &offset, device, &inode, next_path_name) != EOF) {
-      // alignMemAddr aligns the address such that HUGEPAGES/2MB can also be used.
-      prev_addr_end = alignMemAddr(prev_addr_end);
+	if (readMapsLine(mapsfd, &area)) {
+		prev_addr_end = (uint64_t) area.endAddr;
+		mtcp_strcpy(prev_path_name, area.name);
+	}
 
-      // We add a 1GB buffer between the end of the heap or the start of the stack.
-      if (strcmp(prev_path_name, "[heap]") == 0) {
-        prev_addr_end += ONE_GB;
-      }
-      if (strcmp(next_path_name, "[stack]") == 0) {
-        next_addr_start += ONE_GB;
-      }
+	while (readMapsLine(mapsfd, &area)) {
+		next_addr_start = (uint64_t) area.addr;
+		next_addr_end = (uint64_t) area.endAddr;
+		mtcp_strcpy(next_path_name, area.name);
 
-      if (prev_addr_end + *region_size <= next_addr_start) {
-        lh_mem_range->start = (VA)prev_addr_end;
-        lh_mem_range->end =   (VA)prev_addr_end + *region_size;
-        is_set = true;
-        break; 
+    // ROUNDADDRUP aligns the address such that HUGEPAGES/2MB can also be used.
+    prev_addr_end = ROUNDADDRUP(prev_addr_end, 2 * ONEMB);
+
+    // We add a 1GB buffer between the end of the heap or the start of the 
+		// stack.
+    if (mtcp_strcmp(prev_path_name, "[heap]") == 0) {
+      prev_addr_end += ONEGB;
     }
-      prev_addr_end = next_addr_end;
-      strcpy(prev_path_name, next_path_name);
+    if (mtcp_strcmp(next_path_name, "[stack]") == 0) {
+      next_addr_start -= ONEGB;
     }
-    fclose(f);
+
+    if (prev_addr_end + 2 * ONEGB <= next_addr_start) {
+      lh_mem_range->start = (VA)prev_addr_end;
+      lh_mem_range->end =   (VA)prev_addr_end + 2 * ONEGB;
+      is_set = true;
+      break; 
+  	}
+    prev_addr_end = next_addr_end;
+    mtcp_strcpy(prev_path_name, next_path_name);
   }
-  JASSERT(is_set)(JASSERT_ERRNO).Text("No memory region can be found for the lower half");
+	close(mapsfd);
+
+	JASSERT(is_set)(JASSERT_ERRNO)
+		.Text("No memory region can be found for the lower half");
 }
 
 // Sets the address range for the lower half. The lower half gets a fixed
@@ -415,14 +411,10 @@ dynamicMemRangeHelper(MemRange_t *lh_mem_range, const uint64_t *region_size)
 static MemRange_t
 setLhMemRange()
 {
-  const uint64_t ONE_GB = 0x40000000;
-  const uint64_t TWO_GB = 0x80000000;
   Area area;
   bool found = false;
   int mapsfd = open("/proc/self/maps", O_RDONLY);
-  if (mapsfd < 0) {
-    JASSERT(false)(JASSERT_ERRNO).Text("Failed to open proc maps");
-  }
+  JASSERT(mapsfd >= 0)(JASSERT_ERRNO).Text("Failed to open proc maps");
   while (readMapsLine(mapsfd, &area)) {
     if (strstr(area.name, "[stack]")) {
       found = true;
@@ -432,15 +424,7 @@ setLhMemRange()
   close(mapsfd);
   static MemRange_t lh_mem_range;
   if (found) {
-#if defined(CENTOS)
-    dynamicMemRangeHelper(&lh_mem_range, &TWO_GB);
-#elif !defined(USE_MANA_LH_FIXED_ADDRESS)
-    lh_mem_range.start = (VA)area.addr - TWO_GB;
-    lh_mem_range.end = (VA)area.addr - ONE_GB;
-#else
-    lh_mem_range.start = 0x2aab00000000;
-    lh_mem_range.end =   0x2aab00000000 + ONE_GB;
-#endif
+   findLHMemRange(&lh_mem_range);
   } else {
     JASSERT(false).Text("Failed to find [stack] memory segment\n");
   }
--- a/contrib/mpi-proxy-split/split_process.cpp
+++ b/contrib/mpi-proxy-split/split_process.cpp
@@ -47,7 +47,10 @@
 #include "procmapsutils.h"
 #include "util.h"
 #include "dmtcp.h"
-#include "mtcp/mtcp_util.h"
+
+#ifndef CENTOS
+# define CENTOS
+#endif
 
 static unsigned long origPhnum;
 static unsigned long origPhdr;
@@ -363,13 +366,13 @@ findLHMemRange(MemRange_t *lh_mem_range)
 
   if (readMapsLine(mapsfd, &area)) {
     prev_addr_end = (uint64_t) area.endAddr;
-    strcpy(prev_path_name, area.name);
+    strncpy(prev_path_name, area.name, PATH_MAX - 1);
   }
 
   while (readMapsLine(mapsfd, &area)) {
     next_addr_start = (uint64_t) area.addr;
     next_addr_end = (uint64_t) area.endAddr;
-    strcpy(next_path_name, area.name);
+    strncpy(next_path_name, area.name, PATH_MAX - 1);
 
     // ROUNDADDRUP aligns the address such that HUGEPAGES/2MB can also be used.
     prev_addr_end = ROUNDADDRUP(prev_addr_end, 2 * ONEMB);
@@ -418,7 +421,15 @@ setLhMemRange()
   close(mapsfd);
   static MemRange_t lh_mem_range;
   if (found) {
-   findLHMemRange(&lh_mem_range);
+#if defined(CENTOS)
+    findLHMemRange(&lh_mem_range);
+#elif !defined(USE_MANA_LH_FIXED_ADDRESS)
+    lh_mem_range.start = (VA)area.addr - 2 * ONEGB;
+    lh_mem_range.end =   (VA)area.addr - ONEGB;
+#else
+    lh_mem_range.start = 0x2aab00000000;
+    lh_mem_range.end =   0x2aab00000000 + ONEGB;
+#endif
   } else {
     JASSERT(false).Text("Failed to find [stack] memory segment\n");
   }
--- a/contrib/mpi-proxy-split/split_process.cpp
+++ b/contrib/mpi-proxy-split/split_process.cpp
@@ -348,6 +348,13 @@ startProxy()
   return childpid;
 }
 
+// Rounds the given address up to the nearest 2MB.
+static inline unsigned long long
+alignMemAddr(unsigned long addr)
+{
+  return (addr >> 20) + 0x1 << 20;
+}  
+
 // Sets the address range for the lower half. The lower half gets a fixed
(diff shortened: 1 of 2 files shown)


contrib/mpi-proxy-split/split_process.cpp
```diff
@@ -47,6 +47,10 @@
 #include "util.h"
 #include "dmtcp.h"
 
+#ifndef CENTOS
+# define CENTOS
+#endif
```
fyshhh, line 52: WIP: check os and define CENTOS appropriately.

contrib/mpi-proxy-split/split_process.cpp
```diff
@@ -328,6 +332,67 @@ startProxy()
...
+  unsigned long long next_addr_end;
+
+  // FIXME: fopen and fscanf may call malloc, which might call mmap.
+  // For now, we hope it doesn't change the calculation.
+  FILE* f = fopen("/proc/self/maps", "r");
+  if (f) {
+    fscanf(f, "%llx-%llx %s %lx %s %lx %s\n",
+           &prev_addr_start, &prev_addr_end, perms, &offset, device, &inode, prev_path_name);
+    while (fscanf(f, "%llx-%llx %s %lx %s %lx %s\n",
+           &next_addr_start, &next_addr_end, perms, &offset, device, &inode, next_path_name) != EOF) {
```
jungan, line 370: Please check if `fscanf` returns 7, otherwise, some fields may not be correct.

contrib/mpi-proxy-split/split_process.cpp
```diff
@@ -328,6 +332,67 @@ startProxy()
...
+      // alignMemAddr aligns the address such that HUGEPAGES/2MB can also be used.
+      prev_addr_end = alignMemAddr(prev_addr_end);
```
jungan, line 372: Suggest changing the name to `roundup`,

contrib/mpi-proxy-split/split_process.cpp
```diff
@@ -328,6 +332,67 @@ startProxy()
...
+      }
+
+      if (prev_addr_end + *region_size <= next_addr_start) {
+        lh_mem_range->start = (VA)prev_addr_end;
+        lh_mem_range->end =   (VA)prev_addr_end + *region_size;
+        is_set = true;
+        break; 
+    }
+      prev_addr_end = next_addr_end;
+      strcpy(prev_path_name, next_path_name);
```
jungan, line 389: Please don't use `strcpy`, one good practice I found is how git does it. https://github.com/git/git/blob/master/banned.h

contrib/mpi-proxy-split/split_process.cpp
```diff
@@ -328,6 +332,67 @@ startProxy()
...
+           &next_addr_start, &next_addr_end, perms, &offset, device, &inode, next_path_name) != EOF) {
+      // alignMemAddr aligns the address such that HUGEPAGES/2MB can also be used.
+      prev_addr_end = alignMemAddr(prev_addr_end);
+
+      // We add a 1GB buffer between the end of the heap or the start of the stack.
+      if (strcmp(prev_path_name, "[heap]") == 0) {
+        prev_addr_end += ONE_GB;
+      }
+      if (strcmp(next_path_name, "[stack]") == 0) {
+        next_addr_start += ONE_GB;
```
jungan, line 379: should use minus?

contrib/mpi-proxy-split/split_process.cpp
```diff
@@ -328,6 +332,67 @@ startProxy()
...
+    fscanf(f, "%llx-%llx %s %lx %s %lx %s\n",
+           &prev_addr_start, &prev_addr_end, perms, &offset, device, &inode, prev_path_name);
+    while (fscanf(f, "%llx-%llx %s %lx %s %lx %s\n",
+           &next_addr_start, &next_addr_end, perms, &offset, device, &inode, next_path_name) != EOF) {
+      // alignMemAddr aligns the address such that HUGEPAGES/2MB can also be used.
+      prev_addr_end = alignMemAddr(prev_addr_end);
+
+      // We add a 1GB buffer between the end of the heap or the start of the stack.
+      if (strcmp(prev_path_name, "[heap]") == 0) {
+        prev_addr_end += ONE_GB;
```
jungan, line 376: It's confusing to check both `prev` and `next` in this while loop.
I think we only need to check `current`, which is the line just read from `proc/maps`
Whatever you need to do for `current`, will be copied to `prev`.

contrib/mpi-proxy-split/split_process.cpp
```diff
@@ -328,6 +332,67 @@ startProxy()
...
+static inline unsigned long long
+alignMemAddr(unsigned long long addr)
+{
+  return (addr + 0x1fffff) & ~0x1fffff;
+}
+
+// This sets the address range of the lower half by searching through the memory
+// region for the first available free region.
+static void
+dynamicMemRangeHelper(MemRange_t *lh_mem_range, const uint64_t *region_size)
```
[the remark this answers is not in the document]
jungan: Please change the name to really reflect what this function is doing.
`FindLhMemRange`

contrib/mpi-proxy-split/split_process.cpp
```diff
@@ -352,7 +417,9 @@ setLhMemRange()
   close(mapsfd);
   static MemRange_t lh_mem_range;
   if (found) {
-#if !defined(USE_MANA_LH_FIXED_ADDRESS)
+#if defined(CENTOS)
+    dynamicMemRangeHelper(&lh_mem_range, &TWO_GB);
+#elif !defined(USE_MANA_LH_FIXED_ADDRESS)
```
jungan, line 422: Just found MANA already has `readMapsLine`, I don't think you need another function to do the same thing. Why don't you utilize that one?

And you also don't need to define `CENTOS` here. Only `USE_MANA_LH_FIXED_ADDRESS` is good enough.
The function to find the right region should be general.

contrib/mpi-proxy-split/split_process.cpp
fyshhh: good catch!

contrib/mpi-proxy-split/split_process.cpp
fyshhh: good call. i wrote this before i found those functions, so i based it off of some of the code earlier in the same file. i will change it, and it'll make it a lot more readable as well.

contrib/mpi-proxy-split/split_process.cpp
fyshhh: interesting. was your link supposed to have git's own implementation of `strcpy`? just says that it's banned.

since i'm changing the map-reading process to use `mtcp_readmapsline`, i'll also change this to use `mtcp_strcpy`.

contrib/mpi-proxy-split/split_process.cpp
fyshhh: right, i think that would only be a problem if the very first line read is the heap memory region - is that possible?

contrib/mpi-proxy-split/split_process.cpp
```diff
@@ -417,15 +403,7 @@ setLhMemRange()
   close(mapsfd);
   static MemRange_t lh_mem_range;
   if (found) {
-#if defined(CENTOS)
-    dynamicMemRangeHelper(&lh_mem_range, &TWO_GB);
-#elif !defined(USE_MANA_LH_FIXED_ADDRESS)
```
jungan: Did you test your code on cori? I think this is for cori, please don't remove it unless you confirmed it works on cori as well.

contrib/mpi-proxy-split/split_process.cpp
```diff
@@ -332,65 +329,58 @@ startProxy()
...
-        lh_mem_range->end =   (VA)prev_addr_end + *region_size;
-        is_set = true;
-        break; 
+  while (readMapsLine(mapsfd, &area)) {
+    next_addr_start = (uint64_t) area.addr;
+    next_addr_end = (uint64_t) area.endAddr;
+    strcpy(next_path_name, area.name);
+
+    // ROUNDADDRUP aligns the address such that HUGEPAGES/2MB can also be used.
+    prev_addr_end = ROUNDADDRUP(prev_addr_end, 2 * ONEMB);
```
jungan, line 360: Why don't you do it at line 350?

contrib/mpi-proxy-split/split_process.cpp
```diff
@@ -332,65 +329,58 @@ startProxy()
...
+
+    // We add a 1GB buffer between the end of the heap or the start of the 
+    // stack.
+    if (strcmp(prev_path_name, "[heap]") == 0) {
+      prev_addr_end += ONEGB;
```
jungan, line 365: ditto

contrib/mpi-proxy-split/split_process.cpp
```diff
@@ -332,65 +329,58 @@ startProxy()
...
-      // We add a 1GB buffer between the end of the heap or the start of the stack.
-      if (strcmp(prev_path_name, "[heap]") == 0) {
-        prev_addr_end += ONE_GB;
-      }
-      if (strcmp(next_path_name, "[stack]") == 0) {
-        next_addr_start += ONE_GB;
-      }
+  if (readMapsLine(mapsfd, &area)) {
+    prev_addr_end = (uint64_t) area.endAddr;
+    strcpy(prev_path_name, area.name);
```
jungan, line 351: strncpy

contrib/mpi-proxy-split/split_process.cpp
```diff
@@ -332,65 +329,58 @@ startProxy()
...
+  }
```
jungan, line 352: I think we need to handle `else` case.

contrib/mpi-proxy-split/split_process.cpp
```diff
@@ -46,10 +46,7 @@
 #include "procmapsutils.h"
 #include "util.h"
 #include "dmtcp.h"
-
-#ifndef CENTOS
-# define CENTOS
-#endif
+#include "mtcp/mtcp_util.h"
```
jungan, line 49: I don't think you need it?

fyshhh: yes, i was using it before and i removed my changes. forgot to remove this one.

contrib/mpi-proxy-split/split_process.cpp
fyshhh: do we? if nothing is read, then the while loop fails also, then at the very end the assert statement will fail.

contrib/mpi-proxy-split/split_process.cpp
fyshhh: has to be updated every iteration of the loop.

contrib/mpi-proxy-split/split_process.cpp
fyshhh: not yet, but i will send some tests to cori before i merge this in. i think the way it was written before. i will restore this segment.

contrib/mpi-proxy-split/split_process.cpp
jungan: It's more clear to change the variable when you got it, at line 354:
```
prev_addr_end = ROUNDADDRUP(area.endAddr, 2 * ONEMB);
```
At line 359:
```
next_addr_end = ROUNDADDRUP(area.endAddr, 2 * ONEMB);
```

contrib/mpi-proxy-split/split_process.cpp
```diff
@@ -328,21 +332,71 @@ startProxy()
...
+    }
+
+    if (prev_addr_end + 2 * ONEGB <= next_addr_start) {
+      lh_mem_range->start = (VA)prev_addr_end;
+      lh_mem_range->end =   (VA)prev_addr_end + 2 * ONEGB;
+      is_set = true;
+      break; 
+    }
+    prev_addr_end = next_addr_end;
+    strcpy(prev_path_name, next_path_name);
```
jungan, line 381: strncpy

contrib/mpi-proxy-split/split_process.cpp
```diff
@@ -352,12 +406,14 @@ setLhMemRange()
   close(mapsfd);
   static MemRange_t lh_mem_range;
   if (found) {
-#if !defined(USE_MANA_LH_FIXED_ADDRESS)
-    lh_mem_range.start = (VA)area.addr - TWO_GB;
-    lh_mem_range.end = (VA)area.addr - ONE_GB;
+#if defined(CENTOS)
+    findLHMemRange(&lh_mem_range);
```
jungan, line 410: Why add `CENTOS` back? cori has `USE_MANA_LH_FIXED_ADDRESS` defined by default, doesn't it?

contrib/mpi-proxy-split/split_process.cpp
```diff
@@ -328,21 +332,71 @@ startProxy()
...
+  uint64_t prev_addr_end;
+  uint64_t next_addr_start;
+  uint64_t next_addr_end;
+
+  int mapsfd = open("/proc/self/maps", O_RDONLY);
+  JASSERT(mapsfd >= 0)(JASSERT_ERRNO).Text("Failed to open proc maps");
+
+  if (readMapsLine(mapsfd, &area)) {
+    prev_addr_end = (uint64_t) area.endAddr;
+    strncpy(prev_path_name, area.name, PATH_MAX - 1);
```
jungan, line 354: Add `prev_path_name[PATH_MAX - 1] = '\0'`, `prev_path_name` was not initialized, and it may not null-terminated which can cause weird bug.
Although it's unlikely that the name could exceed`PATH_MAX`, please be a safe coder.

contrib/mpi-proxy-split/split_process.cpp
fyshhh: yes, but this was the initial implementation of it. once i've done tests on cori, we'll just have this be the default way the mem range is determined.

geminijun: Yes, so on cori, it will go
```
#else
    lh_mem_range.start = 0x2aab00000000;
    lh_mem_range.end =   0x2aab00000000 + ONEGB;
#endif
```
Correct?
What I meant is you don't need to add `CENTOS`, just change the code for when `USE_MANA_LH_FIXED_ADDRESS` is not defined. That name implies everything, which is better. Other people won't understand what makes `CENTOS` different.
One day if we don't need `USE_MANA_LH_FIXED_ADDRESS` anymore, then we can remove that as well.

jungan: So on cori, it will go
```
#else
    lh_mem_range.start = 0x2aab00000000;
    lh_mem_range.end =   0x2aab00000000 + ONEGB;
#endif
```
Correct?
What I meant is you don't need to add CENTOS, just change the code for when USE_MANA_LH_FIXED_ADDRESS is not defined. That name implies everything, which is better. Other people won't understand what makes CENTOS different.
One day if we don't need USE_MANA_LH_FIXED_ADDRESS anymore, then we can remove that as well.

contrib/mpi-proxy-split/split_process.cpp
```diff
@@ -352,12 +402,17 @@ setLhMemRange()
...
   if (found) {
+    findLHMemRange(&lh_mem_range);
+// This segment is only here until I ascertain whether the generalized code
+// works on Cori too. The above function works on CentOS. - Illio
+#if 0
 #if !defined(USE_MANA_LH_FIXED_ADDRESS)
-    lh_mem_range.start = (VA)area.addr - TWO_GB;
-    lh_mem_range.end = (VA)area.addr - ONE_GB;
+    lh_mem_range.start = (VA)area.addr - 2 * ONEGB;
+    lh_mem_range.end =   (VA)area.addr - ONEGB;
```
JainTwinkle, line 411: could you remove extra space before `(VA)area.addr`?

jungan (approved):

contrib/mpi-proxy-split/split_process.cpp
```diff
@@ -344,21 +344,70 @@ startProxy()
...
+    next_addr_end = ROUNDADDRUP((uint64_t) area.endAddr, 2 * ONEMB);
+    strncpy(next_path_name, area.name, PATH_MAX - 1);
+    next_path_name[PATH_MAX - 1] = '\0';
+
+    // We add a 1GB buffer between the end of the heap or the start of the 
+    // stack.
+    if (strcmp(next_path_name, "[heap]") == 0) {
+      next_addr_end += ONEGB;
+    }
+    if (strcmp(next_path_name, "[stack]") == 0) {
```
jungan, line 382: NIT, `else if`, can save 1 unnecessary comparison when it's `heap`

jungan:
@gc00 The code looks good to me now, and @fyshhh has tested it on both CORI and CentOS 7. Do you agree to merge it? 

This PR is for `feature/centos`, if it has been tested on CORI, I think we can cherry-pick to master as well. But please always use PR to merge to master. Thanks!

gc00:
@jungan , I agree.  Let's 'rebase and merge' this into feature/centos.

Then we can announce this branch to a few friends.  I propose to let the friends try this out over the next week.  If there are no problems, we can create the PR for 'master'.

@leobago and @kellekai,
Would either of you be interested in trying this out from the feature/centos branch?  (Please verify that this PR has been pushed into feature/centos before testing it.)  This branch should assume that you are running on CentOS.  Eventually, the master branch will auto-detect what O/S is being used, but that's not there yet.

kellekai:
@gc00 sure, we can try it. Maybe to assure that we don't do anything wrong, you could prepare a mini tutorial for us, including the steps from compilation to restart for a simple example.

gc00:
@kellekai,
    Good idea.  We'll try to have a tutorial ready for basic testing in the next few days.
    Later (but not now), we plan to add variants of the tutorial for various use cases.
@fyshhh,
    Could you prepare a tutorial with step-by-step instructions of what to do to install and compile on a new CentOS machine, launch on a given MPI target application, checkpoint and restart?  Let's include instructions both for manual checkpoint and periodic checkpoint.  Let's also include a brief pointer to 'man mana' and to DMTCP_ROOT/test/plugin/applic-initiated-ckpt, showing how the target application can directly request a checkpoint (but it should be requested by just one MPI process).
    At each step, there should be both the instructions of what to do, and a quick check to verify if that step was done correctly.
    The tutorial should point to 'man mana' in the manpages, and to the appropriate README in the mpi-proxy-split plugin.
    The tutorial should also say that for a worst-case target application makes very frequent MPI calls, it is possible to see up to 50% runtime overhead in a worst case.  We are working to reduce that runtime overhead.  But for now, the user should measure the runtime overhead that they see for their particular target application.
    When the tutorial is ready, let me know, and I'll do a second draft to polish it further.
    Thanks.

fyshhh:
@gc00
sure, but i'm currently working on #29, which is part of enabling restart on CentOS machines. once that's done, i'll work on doing this.

kellekai:
Dear Gene (and who else it may concern),

I respectfully wanted to ask if there is any update regarding a
step-by-step tutorial for testing MANA on a fresh CentOS system? I am
experiencing some issues and I am wondering if I am doing everything
correctly.

Thank you very much!

Cheers,

Kai

JainTwinkle:
Hi @kellekai,

I think you are looking for https://github.com/mpickpt/mana/blob/feature/centos/doc/centos-tutorial.txt (https://github.com/mpickpt/mana/pull/37).
Please follow the step-by-step tutorial mentioned in the document. And, let us know if you still see any errors.
You can also create a GitHub issue if you are facing some issues with installation. It'll be helpful for the rest community.

Thanks,
Twinkle

P.S.: I'm tagging @fyshhh  and @gc00 for further help. :)

kellekai:
Hi,

great! I will check it out.

Thanks,

Kai

PR #27, opened 2021-11-06
[MANA-43] Fix gethostbyname static again

gc00:
This is needed for MANA on CentOS, so that mana_launch/dmtcp_launch finds `gethostbyname_proxy`.

There are now three commits (used to be two).

1.  Look for gethostbyname_proxy in PATH environment.  This modifies `gethostbyname-static` directory, and I have tested that directory locally.  It now will find gethostbyname_proxy from ~/bin in PATH, even when `gethostbyname_proxy` has been deleted from the `gethostbyname-static` directory.  (I tested with `./a.out localhost` and `./a.out localhost time`.)

2. Add the DMTCP bin directory to the PATH environment.  It adds about 8 new lines of code to `split-process.cpp`.  I tested this commit in a local example inside DMTCP, but not in MANA.  I didn't have time to test this directly in MANA, and I won't have time for that any time soon.  As part of the review, please rebase feature/centos on top of this branch, and <br>&nbsp;&nbsp; *verify that `mana_launch` on CentOS now works in a new build*.

3.  Several bug fixes (This is a new commit, in response the review comments.)

@jungan and @fyshhh , I've now fixed the bugs from PR #27.  Thanks for the excellent comments.  Please review it again now.

@fyshhh , You said that `make -j mana` wasn't automatically compiling the gethostbyname_static directory.  In the interests of timeliness, let's review and push this in.  If you see how to force 'make' to compile the gethostbyname_static directory before me, please generate a PR for it.

### DIFF
--- a/contrib/mpi-proxy-split/lower-half/gethostbyname-static/gethostbyname_static.c
+++ b/contrib/mpi-proxy-split/lower-half/gethostbyname-static/gethostbyname_static.c
@@ -82,19 +82,25 @@ static int error(struct hostent **result) {
 
 
 // For getaddrinfo or gethostbyname_r
-void execvpProxyWithArgv(const char *file, const char *arg) {
-    char *argv[] = {(char *)file, (char *)arg, NULL};
+void execvpProxyWithArgv(const char *file, const char *arg1, const char *arg2) {
+    char *argv[] = {(char *)file, (char *)arg1, (char *)arg2, NULL};
     int rc = execvp("gethostbyname_proxy", argv);
-    // NOTREACHED unless execvp() fails.
+    // NOTREACHED unless execvp() fails.  Try to execvp alternate path to file.
     // Next, look for gethostbyname_proxy in same dir as this executable.
     char buf[10000];
-    readlink("/proc/self/exe", buf, sizeof(buf));
+    int num = readlink("/proc/self/exe", buf, sizeof(buf));
+    if (num == -1) {perror("readlink"); exit(1);}
+    // Verify neither 'buf' nor 'dirname(buf) + "/gethostbyname_proxy"' overflow
+    assert(num + 1 + sizeof("/gethostbyname_proxy") < sizeof(buf));
+    buf[num] = '\0';
     char *dir = dirname(buf);
     assert(strlen(dir) + sizeof("gethostbyname_proxy") < sizeof(buf));
     char *path = strcat(dir, "/gethostbyname_proxy");
     rc = execvp(path, argv);
+    // This is a real error.
     snprintf(buf, sizeof(buf), "execvp(gethostbyname_proxy) for %s", file);
-    if (rc == -1) { perror(buf); }
+    perror(buf);
+    exit(1);
 }
 
 // FIXME:  The GLIBC man page is ambiguous.  Does '*result' need to
@@ -113,7 +119,7 @@ int gethostbyname_r(const char *name,
     close(pipefd[0]);
     dup2(pipefd[1], 1);
     close(pipefd[1]);
-    execvpProxyWithArgv("gethostbyname_r", name);
+    execvpProxyWithArgv("gethostbyname_r", name, NULL);
   } else if (childpid > 0) {
     close(pipefd[1]);
     int rc = readall(pipefd[0], &hostent_result, sizeof(hostent_result));
@@ -177,7 +183,7 @@ int getaddrinfo(const char *__restrict node,
     close(pipefdin[0]);
     dup2(pipefdout[1], 1);
     close(pipefdout[1]);
-    execvpProxyWithArgv("getaddrinfo", service);
+    execvpProxyWithArgv("getaddrinfo", node, service);
   } else if (childpid > 0) {
     if (hints != NULL) {
       writeall(pipefdin[1], hints, sizeof(*hints));
@@ -194,7 +200,6 @@ int getaddrinfo(const char *__restrict node,
     int rc = readall(pipefdout[0], &addrinfo_result, sizeof(addrinfo_result));
     assert(rc == sizeof(addrinfo_result));
     close(pipefdin[0]);
-    assert(rc == sizeof(addrinfo_result));
     int status;
     waitpid(childpid, &status, 0);
     *res = (struct addrinfo *)addrinfo_result.padding;
--- a/contrib/mpi-proxy-split/lower-half/gethostbyname-static/gethostbyname_static.c
+++ b/contrib/mpi-proxy-split/lower-half/gethostbyname-static/gethostbyname_static.c
@@ -13,6 +13,12 @@
  * limitations under the License.
  */
 
+// TESTING:  make
+// USAGE:  Compile gethostbyname*.{c,h} into your statically linked application
+//         It will now call the gethostbyname_r, getaddrinfo defined here
+//         instead of the versions in glibc.  The versions in glibc would
+//         simply report that those functions are not supported statically.
+
 #include <assert.h>
 #include <errno.h>
 #include <libgen.h>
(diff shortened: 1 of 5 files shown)


jungan (commented):
Thanks, @gc00 for the fix! I guess you won't have time to continue working on this, how about you just reply to my comments and let @fyshhh fix them?

@fyshhh please also update the test result in the comment, thanks!

contrib/mpi-proxy-split/lower-half/gethostbyname-static/gethostbyname_static.c
```diff
@@ -166,12 +177,7 @@ int getaddrinfo(const char *__restrict node,
     close(pipefdin[0]);
     dup2(pipefdout[1], 1);
     close(pipefdout[1]);
-    char *argv[] = {"getaddrinfo", (char *)node, (char *)service, NULL};
```
fyshhh, line 169: i think we should leave this line in, and pass `*argv` to `execvpProxyWithArgv` rather than `service`, as i'm experiencing a bug when `node` isn't passed in.

contrib/mpi-proxy-split/lower-half/gethostbyname-static/gethostbyname_static.c
```diff
@@ -166,12 +177,7 @@ int getaddrinfo(const char *__restrict node,
...
-    // argv[2] (hints) is passed in from parent via writeall
-    char buf[10000];
-    char *path = getProxyPath(buf, sizeof(buf));
-    int rc = execvp(path, argv);
-    if (rc == -1) { perror("execvp"); }
+    execvpProxyWithArgv("getaddrinfo", service);
```
fyshhh, line 180: `node` is not passed in when it should be. i'd suggest changing the implementation to this:
```
char *argv[] = { (char *) node, (char *) service, NULL };
execvpProxyWithArgv("getaddrinfo", *argv);
```

contrib/mpi-proxy-split/lower-half/gethostbyname-static/gethostbyname_static.c
```diff
@@ -75,11 +81,20 @@ static int error(struct hostent **result) {
...
-  assert(strlen(dir) + sizeof("gethostbyname_proxy") < buflen);
-  return strcat(dir, "/gethostbyname_proxy");
+// For getaddrinfo or gethostbyname_r
+void execvpProxyWithArgv(const char *file, const char *arg) {
+    char *argv[] = {(char *)file, (char *)arg, NULL};
+    int rc = execvp("gethostbyname_proxy", argv);
+    // NOTREACHED unless execvp() fails.
+    // Next, look for gethostbyname_proxy in same dir as this executable.
+    char buf[10000];
+    readlink("/proc/self/exe", buf, sizeof(buf));
```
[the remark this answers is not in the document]
gc00: Great catch, Jun!  I'm fixing this in my next update to the PR.

contrib/mpi-proxy-split/lower-half/gethostbyname-static/gethostbyname_static.c
```diff
@@ -75,11 +81,20 @@ static int error(struct hostent **result) {
...
+    char *dir = dirname(buf);
+    assert(strlen(dir) + sizeof("gethostbyname_proxy") < sizeof(buf));
```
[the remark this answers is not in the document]
gc00: Doing something similar in the updated version:
> int num = readlink("/proc/self/exe", buf, sizeof(buf));
> // Verify neither 'buf' nor 'dirname(buf) + "/gethostbyname_proxy"' overflow
> assert(num + 1 + sizeof("/gethostbyname_proxy") < sizeof(buf));

contrib/mpi-proxy-split/lower-half/gethostbyname-static/gethostbyname_static.c
```diff
@@ -75,11 +81,20 @@ static int error(struct hostent **result) {
...
+    char *path = strcat(dir, "/gethostbyname_proxy");
+    rc = execvp(path, argv);
+    snprintf(buf, sizeof(buf), "execvp(gethostbyname_proxy) for %s", file);
+    if (rc == -1) { perror(buf); }
```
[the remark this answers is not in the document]
gc00: That's fair.  Fixed in the update.

contrib/mpi-proxy-split/lower-half/gethostbyname-static/gethostbyname_static.c
```diff
@@ -186,6 +192,7 @@ int getaddrinfo(const char *__restrict node,
     close(pipefdin[1]);
     // addrinfo_result is of fixed size, defined in gethostbyname_static.h
     int rc = readall(pipefdout[0], &addrinfo_result, sizeof(addrinfo_result));
+    assert(rc == sizeof(addrinfo_result));
```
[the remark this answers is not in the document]
gc00: Nice.  Thanks for catching this.

contrib/mpi-proxy-split/lower-half/gethostbyname-static/gethostbyname_static.c
gc00: Thanks, @fyshhh.  This is a really important catch.
And then as I looked more closely, I realized that I was also not capturing the `ai_canonname` (canonical hostname) correctly either.  I've now fixed both in the new update..

contrib/mpi-proxy-split/lower-half/gethostbyname-static/gethostbyname_static.c
```diff
@@ -75,11 +81,26 @@ static int error(struct hostent **result) {
...
 
-char *getProxyPath(char *buf, size_t buflen) {
-  readlink("/proc/self/exe", buf, buflen);
-  char *dir = dirname(buf);
-  assert(strlen(dir) + sizeof("gethostbyname_proxy") < buflen);
-  return strcat(dir, "/gethostbyname_proxy");
+// For getaddrinfo or gethostbyname_r
+void execvpProxyWithArgv(const char *file, const char *arg1, const char *arg2) {
+    char *argv[] = {(char *)file, (char *)arg1, (char *)arg2, NULL};
+    int rc = execvp("execvp(gethostbyname_proxy)", argv);
```
jungan, line 87: I just realized that the `file` passed here, is actually `gethostbyname_r` or `getaddrinfo`, which is the argument that will be passed to `gethostbyname_proxy`.
But by convention, I think the first element in the `argv` should be `gethostbyname_proxy`?
> The execv(), execvp(), and execvpe() functions provide an array of pointers to null-terminated strings that represent the argument list available to the new program. The first argument, by convention, should point to the filename associated with the file being executed. The array of pointers must be terminated by a NULL pointer.

And I guess there is a typo? `execvp("gethostbyname_proxy", argv);`

fyshhh: i did some testing, and jun is right. doing `execvp("execvp(gethostbyname_proxy)", argv)` causes `execvp` to fail.

gc00: Hi @fyshhh and @jungan ,
    As usual, you're right again.  It was a silly error on my part.  I've updated the PR to fix this.  I believe the execvp call works correctly now.

fyshhh: great! did some testing on it and seems like it's working properly now.

jungan (approved):

jungan:
test this please

PR #29, opened 2021-11-14
[MANA-59] Implement dynamic memory region allocation for mtcp_restart.c

fyshhh:
This pull request enables `mana_restart` functionalities on CentOS. Notably, it dynamically determines whether there are vvar and vDSO segments that need to be temporarily remapped, and finds regions to map them to. This also removes some of the hard-coding done when reserving memory regions, as well as slightly changing the implementation of determining the regions to reserve.

This has been tested on both Cori and CentOS by running all the programs in `contrib/mpi-proxy-split/test` directories, and the results seen are the same as when running `mana_launch`. I also checked the addresses assigned for specific cases to ensure that all the code is necessary.

Eventually, we should seek to remove all uses of the `GNI` and `MANA_USE_LH_FIXED_ADDRESS` macros, but this is a good stopgap.

### DIFF
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -510,7 +510,8 @@ uint64_t vvarStartTmp = -1;
 uint64_t vdsoStartTmp = -1;
 
 // This function searches for a free memory region to temporarily remap the vdso
-// and vvar regions to, as these regions will be unmapped later on.
+// and vvar regions to, so that mtcp's vdso and vvar do not overlap with the
+// checkpointed process' vdso and vvar.
 static void
 remap_vdso_and_vvar_regions() {
   Area area;
@@ -537,8 +538,8 @@ remap_vdso_and_vvar_regions() {
     mtcp_abort();
   }
 
-  // We have to iterate through the memory map separately to ensure that
-  // temporary regions have been found before we remap.
+  // We have to iterate through the memory map twice, this time to ensure that
+  // free regions have been found for us to remap to.
   mapsfd = mtcp_sys_open2("/proc/self/maps", O_RDONLY);
   if (mapsfd < 0) {
     MTCP_PRINTF("error opening /proc/self/maps; errno: %d\n", mtcp_sys_errno);
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -67,7 +67,7 @@
 #define HUGEPAGES
 #define GNI
 
-/*TODO: @Illio, this macro is used for CentOS specific changes for now.
+/*TODO: This macro is used for CentOS specific changes for now.
  * Later, we want to merge the code that works both for CentOS and Cori. 
  */
 // if we define both GNI and CENTOS, assume we are running on CentOS
@@ -534,22 +534,6 @@ setupTempRegions() {
   }
 }
 
-#if 0
-#define PAGESIZE 4096 /* We hardwire this, since we can't make libc calls. */
-// We choose this address based on our observation on CORI that it won't overlap
-// with any other memory segments. We might change this address in the future,
-// for now, it doesn't matters.
-#define vvarStartTmp (0x40000 - 5*PAGESIZE)
-#define vdsoStartTmp (0x40000 - 2*PAGESIZE)
-/* @Illio, this is hardwired for now and seem to work. Can we figure out the
- * location dynamically? or can we use only one area that we are reserving for
- * LH anyway?
- *
- * As it turns out, any free region is fine, as long as it's PAGESIZE aligned
- * and greater than 0x10000.
- */
-#endif
-
 #define shift argv++; argc--;
 NO_OPTIMIZE
 int
@@ -678,23 +662,6 @@ main(int argc, char *argv[], char **environ)
     //   before vvar and after vdso, and verify that we get an EFAULT.
     // In May, 2020, on Cori and elsewhere, vvar is 3 pages and vdso is 2 pages.
     char *vdsoStart = (char *)mygetauxval(environ, AT_SYSINFO_EHDR);
-    /* @Illio
-     * Problem 1: the vvar location is Cori-specific and assumed to be 3 pages
-     *    lower than vDSO. However, on CentOS VM, we do not see any vvar section
-     *    at all on restart.
-     * Fix: There should not be any assumptions. We should be checking procmaps
-     *      to get the location of both vvar and vdso regions. And, move only if
-     *      they exist in procmaps. This code can be made generic.
-     * Problem 2: These mapped regions will be removed in
-     *    unmap_memory_areas_and_restore_vdso function as mtcp unmap all its
-     *    regions. And, we'll not be able to move it back to original program's
-     *    vdso region.
-     * What-we-do:
-     * 1. move mtcp's vdso and vvar regions temporarily at a temporary location
-     * 2. unmap all mtcp's memory regions
-     * 3. move from temp locations to ckpt'ed process' vdso vvar location.
-     * Fix: We should be bookeeping regions that we don't want MTCP to unmap 
-     */
     setupTempRegions();
     Area area;
     int mapsfd = mtcp_sys_open2("/proc/self/maps", O_RDONLY);
@@ -719,23 +686,6 @@ main(int argc, char *argv[], char **environ)
       }
     }
     mtcp_sys_close(mapsfd);
-
-#if 0
-    char *vvarStart = vdsoStart - 3 * PAGESIZE;
-//    void *rc1 = mtcp_sys_mremap(vvarStart, 3*PAGESIZE, 3*PAGESIZE,
-//		                MREMAP_FIXED | MREMAP_MAYMOVE, vvarStartTmp);
-    void *rc2 = mtcp_sys_mremap(vdsoStart, 2*PAGESIZE, 2*PAGESIZE,
-		                MREMAP_FIXED | MREMAP_MAYMOVE, vdsoStartTmp);
-    if (rc2 == MAP_FAILED) { // vdso segment can be one page or two pages.
-      rc2 = mtcp_sys_mremap(vdsoStart, 1*PAGESIZE, 1*PAGESIZE,
-		            MREMAP_FIXED | MREMAP_MAYMOVE, vdsoStartTmp);
-    }
-    void * rc1 = 0x0; // @Illio: this should go away in the final code
-    if (rc1 == MAP_FAILED || rc2 == MAP_FAILED) {
-      MTCP_PRINTF("mtcp_restart failed: "
-		  "gdb --mpi \"DMTCP_ROOT/bin/mtcp_restart\" to debug.\n");
-    }
-#endif
 
     // Now that we moved vdso/vvar, we need to update the vdso address
     //   in the auxv vector.  This will be needed when the lower half
@@ -830,13 +780,7 @@ main(int argc, char *argv[], char **environ)
 #endif
     typedef int (*getRankFptr_t)(void);
     int rank = -1;
-    /* FIXME: combine GNI load-unload for both centos and Cori
-    * 
-    * @Illio: Currently, the Gni driver is required only for cori. But, we want
-    *   to generalize this code. So, if on some platform, we need to do extra
-    *   stuff for their network driver or any other library. One can simply add
-    *   here.
-    * */
+    /* FIXME: combine GNI load-unload for both centos and Cori */
 #ifndef CENTOS
     beforeLoadingGniDriverBlockAddressRanges(start1, end1, start2, end2);
 #endif
@@ -1492,9 +1436,6 @@ unmap_memory_areas_and_restore_vdso(RestoreInfo *rinfo, LowerHalfInfo_t *lh_info
       DPRINTF("Skipping lower half memory section: %p-%p\n",
               area.addr, area.endAddr);
     } else if (area.addr == vdsoStartTmpCopy) {
-      // FIXME @Illio: we should replace this with the vDSO and vvar address.
-      // Or if we want to generalize then we can call a function that tells if
-      // the region should not be unmapped.
       DPRINTF("Skipping temporary vDSO section: %p-%p\n",
               area.addr, area.endAddr);
     } else if (area.addr == vvarStartTmpCopy) {
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -771,33 +771,13 @@ main(int argc, char *argv[], char **environ)
       //         mtcp_restart is statically linked, and doesn't need it.
       Area heap_area;
       MTCP_ASSERT(getMappedArea(&heap_area, "[heap]") == 1);
-      // FIXME:  choose higher of this
-      //           and lh_info.memRange.end; // 0x2aab00000000
       start1 = max(heap_area.endAddr, lh_info.memRange.end);
-      // FIXME:  choose lower of this and highMemStart
-      //        and make sure it doesn't intersect with lh_info.memRange
       Area stack_area;
       MTCP_ASSERT(getMappedArea(&stack_area, "[stack]") == 1);
       end1 = min(stack_area.endAddr - 4 * GB, highMemStart - 4 * GB);
-      // end1 = highMemStart - 4 * GB;
       start2 = 0;
       end2 = start2;
     }
-# if 0
-    // this is only here for Cori debugging - it will be deleted before this PR
-    // is merged in.
-    // ***** These constants are hardwired for Cori in May, 2020 *****
-    // *****   DELETE THIS WHEN NO LONGER NEEDED FOR DEBUGGING   *****
-    // The dynamic values of start1, end1 above are preferred to this.
-    void *start1 = (const void*)0x2aaaaaaaf000; // End of vdso
-    //  const void *end1 = (const void*)0x2aaaaaaf8000;
-    //  const size_t len1 = end1 - start1;
-    //  const void *start2 = (const void*)0x2aaaaab1b000; // Start of upper half
-    void *end1 = lh_info.memRange.start; //Start of lower half memory
-    size_t len1 = end1 - start1;
-    char *start2 = 0;
-    char *end2 = start2;
-#endif
     typedef int (*getRankFptr_t)(void);
     int rank = -1;
     reserveUpperHalfMemoryRegionsForCkptImgs(start1, end1, start2, end2);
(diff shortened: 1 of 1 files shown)


src/mtcp/mtcp_restart.c
```diff
@@ -632,19 +683,68 @@ main(int argc, char *argv[], char **environ)
     //   before vvar and after vdso, and verify that we get an EFAULT.
     // In May, 2020, on Cori and elsewhere, vvar is 3 pages and vdso is 2 pages.
     char *vdsoStart = (char *)mygetauxval(environ, AT_SYSINFO_EHDR);
```
fyshhh, line 685: i suspect this is now obsolete.

src/mtcp/mtcp_restart.c
```diff
@@ -632,19 +683,68 @@ main(int argc, char *argv[], char **environ)
...
+    void * rc1 = 0x0; // @Illio: this should go away in the final code
     if (rc1 == MAP_FAILED || rc2 == MAP_FAILED) {
       MTCP_PRINTF("mtcp_restart failed: "
 		  "gdb --mpi \"DMTCP_ROOT/bin/mtcp_restart\" to debug.\n");
     }
+#endif
+
     // Now that we moved vdso/vvar, we need to update the vdso address
     //   in the auxv vector.  This will be needed when the lower half
     //   starts up:  initializeLowerHalf() will be called from splitProcess().
```
fyshhh, line 750: this (and line 751, i can't include it in the highlight) should also be obsolete. `mygetauxval` and `mysetauxval` are only called here.

fyshhh:
there are several comments left by twinkle and replies made by me in the code. they're only here while the code is under review - once it is ready for merge, they'll either be removed or succinctly summarized to explain certain design decisions.

src/mtcp/mtcp_restart.c
```diff
@@ -496,13 +502,38 @@ mysetauxval(char **evp, unsigned long int type, unsigned long int val)
...
+    mtcp_abort();
+  }
+
+  while (mtcp_readmapsline(mapsfd, &area)) {
+    if (prev_addr + 5 * PAGESIZE <= area.addr) {
+      vvarStartTmp = prev_addr;
+      vdsoStartTmp = prev_addr + 3 * PAGESIZE;
+      break;
+    }
+    prev_addr = ROUNDADDRUP((uint64_t) area.endAddr, 0x1000);
```
jungan, line 528: use `PAGESIZE`, avoid using magic number as much as possible.

src/mtcp/mtcp_restart.c
```diff
@@ -496,13 +502,38 @@ mysetauxval(char **evp, unsigned long int type, unsigned long int val)
...
+  int mapsfd = mtcp_sys_open2("/proc/self/maps", O_RDONLY);
+  if (mapsfd < 0) {
+    MTCP_PRINTF("error opening /proc/self/maps; errno: %d\n", mtcp_sys_errno);
+    mtcp_abort();
+  }
+
+  while (mtcp_readmapsline(mapsfd, &area)) {
+    if (prev_addr + 5 * PAGESIZE <= area.addr) {
+      vvarStartTmp = prev_addr;
+      vdsoStartTmp = prev_addr + 3 * PAGESIZE;
```
jungan, line 525: Why choose 5 and 3 here? Please add comments to explain that

And if they're always 2 pages apart, you only need 1 variable I think.

src/mtcp/mtcp_restart.c
```diff
@@ -632,19 +663,31 @@ main(int argc, char *argv[], char **environ)
...
+      if (mtcp_strcmp(area.name, "[vdso]") == 0) {
+        rc = mtcp_sys_mremap(area.addr, area.size, area.size,
+                             MREMAP_FIXED | MREMAP_MAYMOVE, vdsoStartTmp);
+      }
+      if (rc == MAP_FAILED) {
+        MTCP_PRINTF("mtcp_restart failed: "
+                    "gdb --mpi \"DMTCP_ROOT/bin/mtcp_restart\" to debug.\n");
+      }
     }
+    mtcp_sys_close(mapsfd);
```
jungan, line 689: First, I think we can use one function to do them all, it doesn't make too much sense to me to separate them.

Second, why do we need to remap the vvar and vdso? Please add comments to explain.

src/mtcp/mtcp_restart.c
```diff
@@ -738,13 +781,18 @@ main(int argc, char *argv[], char **environ)
 #endif
     typedef int (*getRankFptr_t)(void);
     int rank = -1;
+    /* FIXME: combine GNI load-unload for both centos and Cori */
+#ifndef CENTOS
     beforeLoadingGniDriverBlockAddressRanges(start1, end1, start2, end2);
+#endif
```
jungan, line 787: I would suggest changing the design like this:
1. change the function name to `reserve_memory_regions`
2. In the implementation of this function, use `#ifdef` to have different implementations for different cases,
```
#ifdef GNI
// reserve the memroy regions
#endif
// do nothing for other cases
```
This would be much more cleaner and general, and easier to expand if we need to add a new case.
https://en.wikipedia.org/wiki/Open–closed_principle

src/mtcp/mtcp_restart.c
```diff
@@ -738,13 +781,18 @@ main(int argc, char *argv[], char **environ)
...
     JUMP_TO_LOWER_HALF(lh_info.fsaddr);
     // MPI_Init is called here. GNI memory areas will be loaded by MPI_Init.
     rank = ((getRankFptr_t)lh_info.getRankFptr)();
     RETURN_TO_UPPER_HALF();
 
+#ifndef CENTOS
     afterLoadingGniDriverUnblockAddressRanges(start1, end1, start2, end2);
+#endif
```
jungan, line 795: ditto

src/mtcp/mtcp_restart.c
```diff
@@ -67,6 +67,12 @@
 #define HUGEPAGES
 #define GNI
 
+/*TODO: This macro is used for CentOS specific changes for now.
+ * Later, we want to merge the code that works both for CentOS and Cori. 
+ */
+// if we define both GNI and CENTOS, assume we are running on CentOS
+#define CENTOS
```
jungan, line 74: Line 68, I think we shouldn't have `GNI` when building on centos.

src/mtcp/mtcp_restart.c
fyshhh: yes, i was speaking to twinkle about this, and this function should be working on CentOS as well. i'll try to debug it with the goal of generalizing it such that we don't require an `#ifdef` clause.

src/mtcp/mtcp_restart.c
fyshhh: 1. that's true.
2. noted.

src/mtcp/mtcp_restart.c
fyshhh: using two might make it clearer for the reader, especially later on when we are comparing addresses. agreed on the additional comments, though.

src/mtcp/mtcp_restart.c
fyshhh: hmm, what do you think @JainTwinkle?

src/mtcp/mtcp_restart.c
```diff
@@ -519,20 +523,47 @@ setupTempRegions() {
...
   mtcp_sys_close(mapsfd);
 
   if (vvarStartTmp < 0 || vdsoStartTmp < 0) {
     MTCP_PRINTF("No free region found to temporarily map vvar/vdso to\n");
     mtcp_abort();
   }
+
+  // We have to iterate through the memory map twice, this time to ensure that
+  // free regions have been found for us to remap to.
+  mapsfd = mtcp_sys_open2("/proc/self/maps", O_RDONLY);
```
jungan, line 544: One of the reasons why I suggested to combine them in the same function is because I don't think it necessary to re-open the same file and re-iterate again.

src/mtcp/mtcp_restart.c
```diff
@@ -519,20 +523,47 @@ setupTempRegions() {
...
+  while (mtcp_readmapsline(mapsfd, &area)) {
+    void *rc = 0x0;
+    if (mtcp_strcmp(area.name, "[vvar]") == 0) {
+      rc = mtcp_sys_mremap(area.addr, area.size, area.size,
+                           MREMAP_FIXED | MREMAP_MAYMOVE, vvarStartTmp);
+    }
+    if (mtcp_strcmp(area.name, "[vdso]") == 0) {
+      rc = mtcp_sys_mremap(area.addr, area.size, area.size,
+                           MREMAP_FIXED | MREMAP_MAYMOVE, vdsoStartTmp);
+    }
```
jungan, line 559: Can we assume there will be only 1 vvar and 1 vdso? Then we probably want to break the loop earlier.

The entries in the `/proc/self/maps` could be very large if it uses a lot of memory.

src/mtcp/mtcp_restart.c
```diff
@@ -519,20 +523,47 @@ setupTempRegions() {
...
+    if (rc == MAP_FAILED) {
+      MTCP_PRINTF("mtcp_restart failed: "
+                  "gdb --mpi \"DMTCP_ROOT/bin/mtcp_restart\" to debug.\n");
```
jungan, line 562: do we want to abort to return an error here?

src/mtcp/mtcp_restart.c
fyshhh: i have a break condition, but it may not work if there isn't a vvar (like on CentOS).

src/mtcp/mtcp_restart.c
fyshhh: yes, i think we should, but i would defer to the opinions of @gc00 and @JainTwinkle.

contrib/mpi-proxy-split/Makefile
```diff
@@ -108,6 +110,7 @@ install: ${MANA_COORD_OBJS}
 	make ${DMTCP_ROOT}/lib/dmtcp/libmana.so
 	make ${DMTCP_ROOT}/lib/dmtcp/libmpidummy.so
 	make ${DMTCP_ROOT}/bin/lh_proxy
+	make ${DMTCP_ROOT}/bin/gethostbyname_proxy
```
jungan, line 113: Please rebase, don't include unrelated commits.

src/mtcp/mtcp_restart.c
```diff
@@ -496,13 +502,92 @@ mysetauxval(char **evp, unsigned long int type, unsigned long int val)
...
+    mtcp_abort();
+  }
+
+  while (mtcp_readmapsline(mapsfd, &area)) {
+    if (mtcp_strcmp(area.name, "[vvar]") == 0) {
+      vvarStart = area.addr;
+      vvarSize = area.size;
+    }
+
+    if (mtcp_strcmp(area.name, "[vdso]") == 0) {
```
jungan, line 538: if it's `vvar`, it won't be possible to be `vdso`, right? why check it twice?

src/mtcp/mtcp_restart.c
```diff
@@ -496,13 +502,92 @@ mysetauxval(char **evp, unsigned long int type, unsigned long int val)
...
+    if (vvarStart > 0 && vdsoStart > 0 && vvarStartTmp > 0 && vdsoStartTmp > 0)
+    {
+      break;
+    }
+  }
+
+  mtcp_sys_close(mapsfd);
+
+  if ((vvarStart > 0 && vvarStartTmp == 0) ||
+    (vdsoStartTmp > 0 && vdsoStartTmp == 0)) {
```
jungan, line 563: I guess it should be `(vdsoStart > 0 && vdsoStartTmp == 0)` mean?

BTW, you can just check the "temp region" at line 569 and 580.

src/mtcp/mtcp_restart.c
```diff
@@ -496,13 +502,92 @@ mysetauxval(char **evp, unsigned long int type, unsigned long int val)
...
+    MTCP_PRINTF("No free region found to temporarily map vvar/vdso to\n");
+    mtcp_abort();
+  }
+
+  if (vvarStart > 0) {
+    rc = mtcp_sys_mremap(vvarStart, vvarSize, vvarSize,
+                         MREMAP_FIXED | MREMAP_MAYMOVE, vvarStartTmp);
+  }
+
+  if (rc == MAP_FAILED) {
```
jungan, line 573: Usually, we want to check it right after we assign it.

When `vvarStart == 0`, it makes no sense to check it, and easy to break when someone else need to add something in this function.

src/mtcp/mtcp_restart.c
fyshhh: yes, you're right. thanks for catching it.

src/mtcp/mtcp_restart.c
```diff
@@ -672,7 +740,6 @@ main(int argc, char *argv[], char **environ)
     total_reserved_fds = reserve_fds_upper_half(reserved_fds);
 
     // Refer to "blocked memory" in MANA Plugin Documentation for the addresses
-#ifdef GNI
 # define GB (uint64_t)(1024 * 1024 * 1024)
     // FIXME:  Rewrite this logic more carefully.
     char *start1, *start2, *end1, *end2;
```
jungan, line 745: I'm curious when will we go this routine?
`(libsEnd + 1 * GB < highMemStart)`

src/mtcp/mtcp_restart.c
```diff
@@ -705,45 +771,22 @@ main(int argc, char *argv[], char **environ)
...
-      // FIXME:  choose lower of this and highMemStart
-      //        and make sure it doesn't intersect with lh_info.memRange
+      start1 = max(heap_area.endAddr, lh_info.memRange.end);
       Area stack_area;
       MTCP_ASSERT(getMappedArea(&stack_area, "[stack]") == 1);
-      end1 = stack_area.endAddr - 4 * GB;
-      // end1 = highMemStart - 4 * GB;
+      end1 = min(stack_area.endAddr - 4 * GB, highMemStart - 4 * GB);
       start2 = 0;
       end2 = start2;
```
jungan, line 779: It looks like `start2` and `end2` are always 0? Then why do we need them?

src/mtcp/mtcp_restart.c
fyshhh: this happens on Cori.

src/mtcp/mtcp_restart.c
JainTwinkle: @jungan and @fyshhh, I'm not very sure either. But, I guess the intention was to show that if we want to block/reserve more than one region, then you can have another pair of `start` and `end`. This would be helpful if you want to reserve two small memory regions instead of one large region. Or if you have some restrictions in blocking one large region and you have to reserve in parts.
I believe @gc00  and Yao kept this code if at all they need it in the future but never used it.
**The question here is:** 
1. Do we need this code block or not?
2. Can we add a comment to describe the intention/purpose of `start2` and `end2`?

src/mtcp/mtcp_restart.c
```diff
@@ -495,13 +496,90 @@ mysetauxval(char **evp, unsigned long int type, unsigned long int val)
...
+      vvarSize = area.size;
+    } else if (mtcp_strcmp(area.name, "[vdso]") == 0) {
+      vdsoStart = area.addr;
+      vdsoSize = area.size;
+    }
+
+    if (vvarStartTmp == 0) {
+      // In May 2020, on Cori and elsewhere, vvar is 3 pages and vdso is 2
+      // pages.
+      if (prev_addr + 5 * PAGESIZE <= area.addr) {
```
JainTwinkle, line 538: Why are we assuming the `vvar` region's size here? 
The `/proc/maps` should give us the exact size of each region on that platform?

src/mtcp/mtcp_restart.c
jungan: Yeah, if would be helpful if we can have some comments here.

And If we really want to have some flexibility here, I think we can use a list of addresses.

src/mtcp/mtcp_restart.c
fyshhh: hmm, that's true. i think we can try finding the regions and their respective sizes first, then reset the file pointer and look for empty regions to map them too.

src/mtcp/mtcp_restart.c
fyshhh: i think we can consider removing the entire block and maybe think of making it open for extension (like jun's suggestion). this is so that in the future, if we need three, or four, or more small regions, we don't just append them to the arguments. we can also use `stdarg` instead, although that might be less safe.

src/mtcp/mtcp_restart.c
```diff
@@ -495,13 +496,90 @@ mysetauxval(char **evp, unsigned long int type, unsigned long int val)
...
-#define PAGESIZE 4096 /* We hardwire this, since we can't make libc calls. */
-// We choose this address based on our observation on CORI that it won't overlap
-// with any other memory segments. We might change this address in the future,
-// for now, it doesn't matters.
-#define vvarStartTmp ((0x40000) - 5*PAGESIZE)
-#define vdsoStartTmp ((0x40000) - 2*PAGESIZE)
+/* We hardwire these, since we can't make libc calls. */
+#define PAGESIZE 4096
+#define ROUNDADDRUP(addr, size) ((addr + size - 1) & ~(size - 1))
+// I think we're forced to use global variables, since
```
JainTwinkle, line 502: Could you elaborate that why the global variable would persist if the `unmap_memory_areas_and_restore_vdso` function is unmapping all mtcp regions including heap and data sections?

src/mtcp/mtcp_restart.c
JainTwinkle: Yes, I agree with removing these extra arguments. And, I'd suggest either changing it to a list or having a comment that currently, we reserve only one large area. If someone needs to extend it in the future then they can use it as a list (and give an example function signature).
@gc00 any comments or suggestions?

src/mtcp/mtcp_restart.c
fyshhh: i'm not too sure what you mean by this. i'm using global variables here because another function uses it later on.

JainTwinkle: It is a bit informal to say "I think we're forced". I meant to update the comment stating the reason instead of a dubious statement. Rest LGTM :)

JainTwinkle (approved):
Please address the last comment and rest LGTM :)

jungan (approved):

src/mtcp/mtcp_restart.c
fyshhh: ah ok, very true.

PR #30, opened 2021-11-16
[MANA-58] Save lower-half BSS region information if the region exists

JainTwinkle:
The current MANA code assumes that there's no BSS section for `lh_proxy` binary. However, this may not be true for all platforms. The restart will fail if MANA checkpoints on such a platform where the BSS region exists. Because we never saved this information and `mtcp_restart` will unmap this essential region of the lower half on restart.
The commit generalizes the code and works whether the BSS region exists or not.

### DIFF
--- a/contrib/mpi-proxy-split/lower-half/libproxy.c
+++ b/contrib/mpi-proxy-split/lower-half/libproxy.c
@@ -56,7 +56,7 @@ static void* MPI_Fnc_Ptrs[] = {
 // Local functions
 
 static void
-getDataFromMaps(const Area *text, Area *data, Area *heap)
+getDataFromMaps(const Area *text, Area *data, Area *bss, Area *heap)
 {
   Area area;
   int mapsfd = open("/proc/self/maps", O_RDONLY);
@@ -68,12 +68,22 @@ getDataFromMaps(const Area *text, Area *data, Area *heap)
       break;
     }
   }
-  // NOTE: Assume that data and heap are contiguous.
+  // NOTE: Assume that data, bss(if exists), and heap are contiguous.
+  /*
+   * The BSS section can be present on some platform and abscent on some.
+   * For e.g., the BSS section is abscent on Cori but present on SLES VM.
+   *
+   * The following code stops after heap region and saves BSS section address
+   * if it exists.
+   */
   void *heap_sbrk = sbrk(0);
   while (readMapsLine(mapsfd, &area)) {
     if (strstr(area.name, "[heap]") && area.endAddr >= (VA)heap_sbrk) {
       *heap = area;
       break;
+    } else {
+	// This should be the BSS section.
+        *bss = area;
     }
   }
   close(mapsfd);
@@ -212,7 +222,7 @@ void first_constructor()
     unsigned long start, end, stackstart;
     unsigned long pstart, pend, pstackstart;
     unsigned long fsaddr = 0;
-    Area txt, data, heap;
+    Area txt, data, bss, heap;
     getTextSegmentRange(getpid(), &start, &end, &stackstart);
     getTextSegmentRange(getppid(), &pstart, &pend, &pstackstart);
     syscall(SYS_arch_prctl, ARCH_GET_FS, &fsaddr);
@@ -220,7 +230,9 @@ void first_constructor()
     end   = ROUND_UP(end);
     txt.addr = (VA)start;
     txt.endAddr = (VA)end;
-    getDataFromMaps(&txt, &data, &heap);
+    // We assign bss to NULL to avoid garbage value if there's no BSS region.
+    bss.addr = NULL;
+    getDataFromMaps(&txt, &data, &bss, &heap);
 
     // TODO: Verify that this gives us the right value every time
     // Perhaps use proc maps in the future?
@@ -230,6 +242,7 @@ void first_constructor()
     lh_info.startText = (void*)start;
     lh_info.endText = (void*)end;
     lh_info.startData = (void*)data.addr;
+    lh_info.startBSS = (void*)bss.addr;
     lh_info.endOfHeap = (void*)heap.endAddr;
     lh_info.libc_start_main = &__libc_start_main;
     lh_info.main = &main;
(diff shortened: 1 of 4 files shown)


contrib/mpi-proxy-split/lower-half/libproxy.c
```diff
@@ -68,12 +68,22 @@ getDataFromMaps(const Area *text, Area *data, Area *heap)
       break;
     }
   }
-  // NOTE: Assume that data and heap are contiguous.
+  // NOTE: Assume that data, bss(if exists), and heap are contiguous.
+  /*
+   * The BSS section can be present on some platform and abscent on some.
```
rohgarg, line 73: abscent -> absent on others.

contrib/mpi-proxy-split/lower-half/libproxy.c
```diff
@@ -68,12 +68,22 @@ getDataFromMaps(const Area *text, Area *data, Area *heap)
...
+   * For e.g., the BSS section is abscent on Cori but present on SLES VM.
```
rohgarg, line 74: abscent -> absent

contrib/mpi-proxy-split/lower-half/libproxy.c
```diff
@@ -68,12 +68,22 @@ getDataFromMaps(const Area *text, Area *data, Area *heap)
...
+   * The following code stops after heap region and saves BSS section address
+   * if it exists.
+   */
   void *heap_sbrk = sbrk(0);
   while (readMapsLine(mapsfd, &area)) {
     if (strstr(area.name, "[heap]") && area.endAddr >= (VA)heap_sbrk) {
       *heap = area;
       break;
+    } else {
+	// This should be the BSS section.
```
rohgarg, line 85: Fix comment alignment

contrib/mpi-proxy-split/lower-half/libproxy.c
```diff
@@ -68,12 +68,22 @@ getDataFromMaps(const Area *text, Area *data, Area *heap)
...
+        *bss = area;
```
rohgarg, line 86: Don't need to break here?

src/mtcp/mtcp_restart.c
```diff
@@ -1305,7 +1305,8 @@ skip_lh_memory_region_ckpting(const Area *area, LowerHalfInfo_t *lh_info)
       mtcp_strstr(area->name, "/dev/xpmem") ||
       mtcp_strstr(area->name, "/dev/shm") ||
       mtcp_strstr(area->name, "/SYS") ||
-      area->addr == lh_info->startData) {
+      area->addr == lh_info->startData ||
+      ((lh_info->startBSS != NULL) && area->addr == lh_info->startBSS)) {
```
rohgarg, line 1309: Maybe move this condition to a const bool above line 1302.

```C
const bool is_lh_bss_region = (lh_info->startBSS != NULL) && area->addr == lh_info->startBSS);
```

Then, you can reference the bool variable here. It'll improve code readability.

contrib/mpi-proxy-split/lower-half/libproxy.c
rohgarg: Also, can you update the comment and add a picture of the memory layout?

contrib/mpi-proxy-split/lower-half/libproxy.c
rohgarg: Is there a way to validate this assumption in the code that the segment following the data segment is the BSS? I mean instead of a blanket "else" case, can we add some additional check? This will save us a lot of debugging pain later if this assumption is violated on a different platform.

rohgarg (changes requested):
General comments:
 - Can we change the name of `skip_lh_memory_region_ckpting()` to `skip_lh_memory_region_unmapping()`? This will make it more meaningful.
 - Change the title of the commit message to: "Don't unmap lower-half BSS region during restart".
 - Also, update the commit message to reflect this. Specifically, this has nothing to do with the platform where we checkpointed. The issue, if I understand it correctly, is about accidentally unmapping a relevant lower-half memory region during restart.

contrib/mpi-proxy-split/lower-half/libproxy.c
gc00: Responding to @rohgarg :
> Don't need to break here? 
The `break` is required.  The logic is to set the bss section to the section just before the 'heap' section is found.  Please keep the 'break'.

contrib/mpi-proxy-split/lower-half/libproxy.c
gc00: For the picture, it's enough to give an example /proc/*/maps for a.out.  But can't we have four memory segments in the worst case, as below?
> ... r-x .../a.out # text segment
> ... r-- .../a.out  # read-only data segment
> ... rw- .../a.out # data segment
> ... rw- .../a.out # bss segment

contrib/mpi-proxy-split/lower-half/libproxy.c
gc00: @rohgarg asks if we can add a check that we've found the bss section, to save possible painful debugging in the future if we fail to identify the bss section.  I agree.  Here is a possible suggestion in that direction.
    First, let's add a comment that we assume that prior to the heap, there are at most three segments (text, data, and bss)., or maybe also a read-only data segment   I suggest a stronger check as follows.

In the `txt` variable of type `Area`, after setting `txt.start = ...`, we can do:
1. Change `getTextSegmentRange()` to also save as an output parameter the (up to) 6 characters of `cmd` (aka `comm` in `man 5 proc`) in that `fscanf`.  (saved from /proc/*/stat; see `man 5 proc`)
2. After `txt.endAddr = (VA)end;`, add:
`strcpy(txt.name, cmd)` (but remove the open and close parentheses)
3. In the `if`, add `else if (strstr(area.name, txt.name)` and do `executableCounter++`.  (and initialize `executableCounter = 0`)
4. In the `else`, do `executableCounter = 0`
5. In the `if` condition that verifies that `heap` was seen, assert that `executableCounter >= 2` and `executableCounter <= 4`.

gc00:
@JainTwinkle , Could you take another look at this PR #30.  We had only some simple changes requested in the PR, and then it was good to go.

Could you make those simple changes, and then ask for a new review?

Thanks very much.

JainTwinkle:
This task is already handled by another PR. So, I'm closing this one.

PR #35, opened 2021-11-30
Misc changes

jungan:
This PR does:
1. Make the "memory access tool" can be used in C++
2. Fix all the warnings, no warnings now.
3. Use MPI_LOGGING instead of LOGGING to avoid conflict

### DIFF
--- a/contrib/mpi-proxy-split/record-replay.h
+++ b/contrib/mpi-proxy-split/record-replay.h
@@ -59,7 +59,8 @@
 // Returns true if we are currently replaying the MPI calls from the saved MPI
 // calls log; false, otherwise. Normally, this would be true while restoring
 // the MPI state at restart time. All other times, this would return false.
-#define LOGGING() \
+// We cannot use LOGGING since it's used for enabling JTRACE
+#define MPI_LOGGING() \
   dmtcp_mpi::MpiRecordReplay::instance().isReplayOn()
 
 // Calls the wrapper function corresponding to the given type 'type'. (The
--- a/src/processinfo.h
+++ b/src/processinfo.h
@@ -149,8 +149,8 @@ class ProcessInfo
 
     uint64_t endOfStack(void) const { return _endOfStack; }
 
-    bool vdsoOffsetMismatch(ptrdiff_t f1, ptrdiff_t f2,
-                            ptrdiff_t f3, ptrdiff_t f4);
+    bool vdsoOffsetMismatch(uint64_t f1, uint64_t f2,
+                            uint64_t f3, uint64_t f4);
 
     string getCkptFilename() const { return _ckptFileName; }
 
--- a/contrib/mpi-proxy-split/.gitignore
+++ b/contrib/mpi-proxy-split/.gitignore
@@ -3,3 +3,4 @@ unit-test/*.exe
 proxy
 mpi-wrappers/*.c*
 mpi-wrappers/*.so*
+Makefile_config
--- a/contrib/mpi-proxy-split/lower-half/.gitignore
+++ b/contrib/mpi-proxy-split/lower-half/.gitignore
@@ -1,0 +1,1 @@
+lh_proxy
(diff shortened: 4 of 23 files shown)


contrib/mpi-proxy-split/record-replay.h
```diff
@@ -59,7 +59,7 @@
 // Returns true if we are currently replaying the MPI calls from the saved MPI
 // calls log; false, otherwise. Normally, this would be true while restoring
 // the MPI state at restart time. All other times, this would return false.
-#define LOGGING() \
+#define MPI_LOGGING() \
```
JainTwinkle, line 62: Could you add a comment before macro stating that how `LOGGING` is conflicting with MPI's logging? And, what are some side effects of keeping the logging macro as `LOGGING`?

src/processinfo.h
```diff
@@ -216,10 +216,10 @@ class ProcessInfo
     uint64_t _vvarEnd;
     uint64_t _endOfStack;
 
-    uint64_t _clock_gettime_offset;
-    uint64_t _getcpu_offset;
-    uint64_t _gettimeofday_offset;
-    uint64_t _time_offset;
+    ptrdiff_t _clock_gettime_offset;
```
JainTwinkle, line 219: Why do we need a signed data type (`ptrdiff_t`) here?

contrib/mpi-proxy-split/record-replay.h
jungan: Sure, will do.
Currently if we wanted to enable `JTRACE`, we have to `./configure --enable-logging`, and it won't compile since makefile passed `-DLOGGING` when compiling, and it's conflict with the definition here.

src/processinfo.h
jungan: CMIIWW,

The function calculate this is `dmtcp_dlsym_lib_fnc_offset`, which could be negative offset I think.

I found this from compile warning, it was also used to compare with a `ptrdiff_t` type variable.

JainTwinkle: The function's offset from the base in the text section would be positive only. You can visualize it like below:
.text base addr|------------| 0x100000
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;|&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;|
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;|&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;|
function1_addr|------------| 0x100050
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;|&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;|
function2_addr|------------| 0x100080

However, what you are saying is correct that in both places the data type should be consistent. DMTCP's master branch fixes this compiler warning here: https://github.com/dmtcp/dmtcp/commit/c3a6b100315051713f28fa716c53abc16d13fb1f
Following this commit, we'll be aligned to upstream DMTCP.

jungan: OK, then I will just cherry-pick that one, thanks!

JainTwinkle (approved):
LGTM :)

PR #36, opened 2021-12-01
[MANA-73] Fix Makefile behavior to copy gethostbyname_proxy to bin

fyshhh:
The current implementation does not copy `gethostbyname_proxy` to the `bin` directory as it's part of the `lh_proxy` rule, so if `lh_proxy` is found in the `bin` directory, then `gethostbyname_proxy` will not be copied out. This can be fixed by simply forcing a `gethostbyname_proxy` check.

### DIFF
--- a/contrib/mpi-proxy-split/lower-half/gethostbyname-static/Makefile
+++ b/contrib/mpi-proxy-split/lower-half/gethostbyname-static/Makefile
@@ -23,3 +23,8 @@ clean:
 dist: clean
 	dir=`basename $$PWD` && cd .. && tar zcvf ./$$dir.tar.gz ./$$dir
 	dir=`basename $$PWD` && ls -l ../$$dir.tar.gz
+
+install:
+	make gethostbyname_proxy
+	cp gethostbyname_proxy -f ../../../../bin
+
--- a/contrib/mpi-proxy-split/Makefile
+++ b/contrib/mpi-proxy-split/Makefile
@@ -98,6 +98,8 @@ ${DMTCP_ROOT}/lib/dmtcp/libmpidummy.so:
 	make -C ${WRAPPERS_SRCDIR} install
 ${DMTCP_ROOT}/bin/lh_proxy:
 	make -C ${LOWER_HALF_SRCDIR} install
+${DMTCP_ROOT}/bin/gethostbyname_proxy:
+	make -C ${LOWER_HALF_SRCDIR}/gethostbyname-static install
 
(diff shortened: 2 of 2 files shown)


JainTwinkle (approved):
LGTM :)

contrib/mpi-proxy-split/lower-half/gethostbyname-static/Makefile
```diff
@@ -23,3 +23,8 @@ clean:
 dist: clean
 	dir=`basename $$PWD` && cd .. && tar zcvf ./$$dir.tar.gz ./$$dir
 	dir=`basename $$PWD` && ls -l ../$$dir.tar.gz
+
+install:
+	make gethostbyname_proxy
+	cp gethostbyname_proxy -f ../../../../bin
```
jungan, line 29: I think this is not a general practice for `make install`
`make install` usually should copy those binaries to the system directory, by default is `/usr/local`, but could be configured by `--prefix` when `./configure`

It looks like other binaries for MANA, `lh_proxy` for example, also has the same problem, I think we can have a separate ticket to fix all of these.

And I would suggest don't use `make install` for development purpose, it looks like it will copy those `dmtcp` binaries to the system directory, which I don't think it's something you want to do in you development.

fyshhh: > And I would suggest don't use make install for development purpose, it looks like it will copy those dmtcp binaries to the system directory, which I don't think it's something you want to do in you development.

I know what you mean by this, but it would seem like `make mana` calls a lot of `make install` instances in subdirectories that doesn't actually copy it to the system directory. This might be a misnomer - perhaps we could recategorize these instances of `install` as something else - maybe `mana`?

jungan: Yes, I found `make mana` will `cd contrib/mpi-proxy-split && make install`, but that is not a right way to do. I think there is a common understanding of `make install` and we should follow the standard.

I'm fine with current version, as long as you know you only need to do `make mana`. And `make install` won't copy mana binaries to the right place, which I filed a ticket to track this issue.

If you check `dmtcp`, you will find that all dmptcp binaries will be generated in the `./bin` directory by just typing `make`, and it won't do `make install`, executing `make install` will copy them to the system directory(`/usr/local` by default, and configurable by `--prefix`) That's the way we should follow.

fyshhh: okay. if there are no further issues, i'll merge this in.

PR #37, opened 2021-12-02
Tutorial for compiling and launching MANA on CentOS machines

fyshhh:
This is a first draft of a tutorial for compiling and launching MANA on CentOS machines. I've verified the correctness of this tutorial by replicating what I've done on a completely new CentOS virtual machine to compile and launch MANA successfully.


doc/centos-tutorial.txt
```diff
@@ -0,0 +1,141 @@
...
+1. Installing essential packages - this section can be skipped depending on the
+existing packages on your machine.
+
+  # yum install -y git
+  # yum groupinstall 'Development Tools'
+  # yum install -y mlocate vim wget
+  # yum install -y centos-release-scl
+  # yum install -y devtoolset-8-gcc devtoolset-8-gcc-c++
+  # echo 'source scl_source enable devtoolset-8' > ~/.bashrc
+  # source ~/.bashrc
```
gc00, line 19: This next section is about using an available MPI, such as MPICH.
Let's add a marker for this section like:
### INSTALLING MPI ###

Also, let's itemize what the MPI package provides, so that we can make MANA more independent of a particular MPI implementation such as MPICH.  Probably, we need 'mpicc' and 'mpic++' and 'libmpi.so' and 'libmpi.a' and 'mpi.h'.  What are the locations assumed by MANA?  Maybe mpicc and mpic++ in the user's path, and libmpi.so/libmp.a/mpi.h in the standard lib and include locations.  If the user has installed libmpi.so/libmp.a/mpi.h, what should the user change in MANA (or change by setting environment variables for LD_LIBRARY_PATH, etc.)?  (For example, it would be nice to make this compatible with a new experimental MPI).
    For this first draft, this information doesn't have to be complete, but let's point to some common choices here.

doc/centos-tutorial.txt
```diff
@@ -0,0 +1,141 @@
...
+
+B. LAUNCHING, CHECKPOINTING AND RESTARTING A MPI APPLICATION
+
+We'll first add the bin directory to PATH. Ensure that these commands are done
+from within the root directory of MANA.
+
+  $ echo 'export PATH=$PATH:{PATH TO MANA}/bin' >> ~/.bashrc
+  $ source ~/.bashrc
+
+By default, this should be /home/{YOUR USERNAME}/mana.
```
gc00, line 65: Unclear.  Are you saying that by default, `{PATH TO MANA}` is `/home/{YOUR USERNAME}/mana`?  Since the user is "ensuring that these commands are done", I don't understand "By default, this should be ...".  Is it about a default in how "the user ensures that these commands are done"?  Perhaps you intended to use a phrase such as "Typically,  `{PATH TO MANA}` will be ..."./

doc/centos-tutorial.txt
```diff
@@ -0,0 +1,141 @@
...
+  # cp ./src/liblzma/.libs/liblzma.a /usr/lib64/
+  # cd ~
+
+2. Installing the MANA directory and compiling MANA
+
+  $ git clone https://github.com/mpickpt/mana.git
+  $ cd mana
+  $ git checkout feature/centos
+  
+Here, we need to make some changes to the configure-mana file - these changes
```
gc00, line 47: Change '-' to '--'.

doc/centos-tutorial.txt
```diff
@@ -0,0 +1,141 @@
...
+  $ echo 'export PATH=$PATH:{PATH TO MANA}/bin' >> ~/.bashrc
+  $ source ~/.bashrc
+
+By default, this should be /home/{YOUR USERNAME}/mana.
+
+1. Launching an MPI application
+
+The MANA directory comes with many test MPI applications that can be found in
+contrib/mpi-proxy-plugin/test. Depending on the application, we may require
+more than 1 copy of the process running - for example, ping_pong.mana.exe
```
gc00, line 71: Change '-' to '--'

doc/centos-tutorial.txt
```diff
@@ -0,0 +1,141 @@
...
+requires 2. To support this, change the argument after -np accordingly. For this
+tutorial, we'll use mpi_hello_world.mana.exe, which only requires 1 copy
```
gc00, line 73: CHANGE: '1' -> 'one'

doc/centos-tutorial.txt
```diff
@@ -0,0 +1,141 @@
...
+
+If the application is launched properly, you should see the following printed:
+
+  Hello world from processor test, rank 0 out of 1 processors
+  Will now sleep for 500 seconds ...
+
+From here, we can either exit the program (through CTRL-C), or continue on to
+the next set of instructions to checkpoint and restart the application.
+
+2. Checkpointing and restarting a MPI application
```
gc00, line 88: Let's split this into `2. Checkpointing an MPI application` and `3. Restarting an MPI application`, and have two seaprate sections.

doc/centos-tutorial.txt
```diff
@@ -0,0 +1,141 @@
...
+background process (note the &).
+
+  $ mana_coordinator
+  $ mpirun -np 1 mana_launch \ 
+      contrib/mpi-proxy-split/test/mpi_hello_world.mana.exe &
+
+Subsequently, we only need the following command to checkpoint the application.
+
+  $ mana_status --checkpoint
+
```
gc00, line 115: `3. Restarting an MPI application` begins here.

doc/centos-tutorial.txt
```diff
@@ -0,0 +1,141 @@
...
+Otherwise, we'll launch an application the same way as above, but as a
+background process (note the &).
+
+  $ mana_coordinator
+  $ mpirun -np 1 mana_launch \ 
+      contrib/mpi-proxy-split/test/mpi_hello_world.mana.exe &
+
+Subsequently, we only need the following command to checkpoint the application.
+
+  $ mana_status --checkpoint
```
gc00, line 114: Let's add a statement here that there are two other ways to specify when to checkpoint.  See `man mana` for details, where the manpage for MANA can be found in `DMTCP_ROOT/manpages`.  The second way is interval checkpointing (`-i`) and the third way is to modify the MPI application so that one and only one MPI process uses the mechanism outlined in `DMTCP_ROOT/test/plugin/applic-initiated-ckpt`.

doc/centos-tutorial.txt
```diff
@@ -0,0 +1,141 @@
...
+
+To restart the application, we do the following:
+
+  $ mana_coordinator
+  $ mpirun -np 1 mana_restart
```
gc00, line 119: Here, it's good to point out that the `mana_restart` command assumes that the checkpoint images are avaialble in the current directory (but write that part more precisely), and that the XXX flag of `mana_restart` allows one to specify a different checkpoint directory.  And that one should see `man mana` for the optional flags to `mana_restart`.

gc00:
@fyshhh, Thanks!  This is a great first start.  Please seem my inline comments for suggestions on how to improve this.

doc/centos-tutorial.txt
```diff
@@ -0,0 +1,141 @@
...
+    [Rank: 0] Choosing ckpt image: ./ckpt_rank_0/ckpt_mpi_hello_world.mana.exe_
+      7c4af7b7e161f9fd-40000-1000f911e89fa5.dmtcp
+  Signal received; continuing sleep for 294 seconds.
+
+C. MISCELLANEOUS RESOURCES
+
+For more information on MANA, you may refer to the manpage in the directory.
+From the root MANA directory, do:
+
+  $ nroff -man manpages/mana.1
```
gc00, line 140: And maybe add:

>D. CAVEATS
At the time of this writing, runtime overhead can be significant for MPI applications that make very frequent calls to MPI.  This is a known issue, and future versions will have lower runtime overhead.

doc/centos-tutorial.txt
fyshhh: you are correct, this is phrased awkwardly.

doc/centos-tutorial.txt
fyshhh: is there a resource that provides a list of the libraries necessary for MANA?

gc00: As we discussed, it's enough if we list the libraries that we are aware of.  No need to spend a lot of time with a deep analysis to list _all_ libraries.  We can extend the tutorial later, based on comments from newer people who try out the tutorial.

gc00 (approved):
@fyshhh , I'm approving the PR now, so that we don't have to wait for the last change to be implemented.  As you see, I'm only asking for a list of obvious libraries that are needed by MANA.  If we can give that list to the user, then they can immediately check if they have the prerequisites (as a best guess), and start to compile.Otherwise, they try to build MANA, and 15 minutes later discover that there's a cryptic message about some missing library.

Just list the obvious libraries that you know we need (e.g., for the static build).  Over time, we will improve this tutorial based on feedback from users.  Thanks.

PR #39, opened 2021-12-03
[MANA-74] Added support for FS base switch in user-space

JainTwinkle:
Get and set FS base register in user-space for kernel version >= 5.3.
This enables MANA to switch to either half in user-space.

### DIFF
--- a/contrib/mpi-proxy-split/split_process.cpp
+++ b/contrib/mpi-proxy-split/split_process.cpp
@@ -24,6 +24,7 @@
   #define _GNU_SOURCE
 #endif
 
+#include <linux/version.h>
 #include <asm/prctl.h>
 #include <sys/prctl.h>
 #include <sys/personality.h>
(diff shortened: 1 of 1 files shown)


contrib/mpi-proxy-split/split_process.cpp
```diff
@@ -70,19 +71,46 @@ static char* proxyAddrInApp();
...
+static inline void setFS(unsigned long fsbase)
+{
+  asm volatile("wrfsbase %0" :: "r" (fsbase) : "memory");
+}
+
 SwitchContext::SwitchContext(unsigned long lowerHalfFs)
 {
   this->lowerHalfFs = lowerHalfFs;
+#if LINUX_VERSION_CODE <= KERNEL_VERSION(5,3,0)
   JWARNING(syscall(SYS_arch_prctl, ARCH_GET_FS, &this->upperHalfFs) == 0)
```
rohgarg, line 96: Why not move this into `setFS()` and `getFS()`? This would make the code to set and get FS in one place. Then you can keep the constructor and destructor simple.

rohgarg (changes requested):

contrib/mpi-proxy-split/split_process.cpp
JainTwinkle: Good point! will do. :)

rohgarg (commented):
Three other general comments:
 - With this change, we cannot take the binaries (and checkpoint image) compiled for > 5.3.0 kernel to an older kernel. Please add this note to the commit message and in the comments. Maybe add an assert, if possible, when the assembly instruction fails. Or will it simply crash on an older kernel? A nicer failure message would be useful, if possible.
 -  Don't you need to do the same thing in `mtcp_split_process.h`? Otherwise, `mtcp_restart` will continue to use the system calls for changing FS base register.
 - Why did you make the functions (`setFS()` and `getFS()`) outside of the class? Unless you intend to use them outside of the class, I'd suggest moving the two functions into the class as private methods. If you intend to use the two functions outside of the class directly, then make them public static methods.

gc00:
This PR (including your commit) has now been incorporated into PR #82.  @JainTwinkle, it should be safe for you to close this PR now.

JainTwinkle:
Closing this PR as the new PR in DMTCP covers the work of this PR.

PR #40, opened 2021-12-16
Merge feature/centos into master branch

fyshhh:
This pull request comprises a few commits made to the `feature/centos` branch to support building, launching, checkpointing and restarting MANA on CentOS, as well as some code generalizations. These commits have been tested on Cori and affirmed to work, by comparing the results from `contrib/mpi-proxy-split/test` with and without these commits.

### DIFF
--- a/configure-mana
+++ b/configure-mana
@@ -9,18 +9,21 @@
 MPI_ETHERNET_INTERFACE=\
 `ip addr |grep -B1 link/ether | head -1 |sed -e 's%[^ ]*: \([^ ]*\): .*%\1%'`
 
-# This works on our local test computer (Ubuntu-18.04).
+# Note: modify this path according to your environment.
+MPI_INSTALL_DIR=/usr/local
+
+# This works on our local test computer (CentOS 7).
 # But also read the note right after this command.
 ./configure --enable-debug \
              CFLAGS=-fno-stack-protector \
              CXXFLAGS=-fno-stack-protector \
-             MPI_BIN=$PWD/../mpich-3.3.2-build/bin \
-             MPI_INCLUDE=$PWD/../mpich-3.3.2-build/include \
-             MPI_LIB=$PWD/../mpich-3.3.2-build/lib \
+             MPI_BIN=$MPI_INSTALL_DIR/bin \
+             MPI_INCLUDE=$MPI_INSTALL_DIR/include \
+             MPI_LIB=$MPI_INSTALL_DIR/lib \
              MPICC='${MPI_BIN}/mpicc' \
-             MPICXX='${MPI_BIN}/mpicxx' \
+             MPICXX='${MPI_BIN}/mpic++' \
              MPIRUN='${MPI_BIN}/mpirun -iface '${MPI_ETHERNET_INTERFACE} \
-             MPI_LD_FLAG=-lmpi \
+             MPI_LD_FLAG=-lmpich \
              MPI_CFLAGS= \
              MPI_CXXFLAGS= \
              MPI_LDFLAGS= \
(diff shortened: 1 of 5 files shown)


configure-mana
```diff
@@ -9,18 +9,21 @@
 MPI_ETHERNET_INTERFACE=\
 `ip addr |grep -B1 link/ether | head -1 |sed -e 's%[^ ]*: \([^ ]*\): .*%\1%'`
 
-# This works on our local test computer (Ubuntu-18.04).
+# Note: modify this path according to your environment.
+MPI_INSTALL_DIR=/usr/local
```
jungan, line 13: Suggest check `MPI_INSTALL_DIR` and only assign default value when it was not set.

BTW, do we use `configure-mana` on CORI?

contrib/mpi-proxy-split/split_process.cpp
```diff
@@ -344,21 +344,69 @@ startProxy()
...
+    }
+
+    if (prev_addr_end + 2 * ONEGB <= next_addr_start) {
+      lh_mem_range->start = (VA)prev_addr_end;
+      lh_mem_range->end = (VA)prev_addr_end + 2 * ONEGB;
+      is_set = true;
+      break; 
+    }
+    prev_addr_end = next_addr_end;
+    strncpy(prev_path_name, next_path_name, PATH_MAX - 1);
```
jungan, line 392: Why do we need to copy `prev_path_name`?

include/procmapsarea.h
```diff
@@ -53,6 +53,11 @@
 #define DELETED_FILE_SUFFIX " (deleted)"
 
 #define FILENAMESIZE        1024
+#define ONEMB (uint64_t)(1024 * 1024)
+#define ONEGB (uint64_t)(1024 * 1024 * 1024)
```
jungan, line 57: I don't think we need `uint64_t` here.

src/mtcp/mtcp_restart.c
```diff
@@ -233,8 +234,8 @@ beforeLoadingGniDriverBlockAddressRanges(char *start1, char *end1,
 }
 
 static void
-afterLoadingGniDriverUnblockAddressRanges(char *start1, char *end1,
-	                                  char *start2, char *end2)
+unreserveUpperHalfMemoryRegionsForCkptImgs(char *start1, char *end1,
```
jungan, line 237: I think `release` is better than `unreserve`.

src/mtcp/mtcp_restart.c
```diff
@@ -495,13 +496,92 @@ mysetauxval(char **evp, unsigned long int type, unsigned long int val)
...
-#define vdsoStartTmp ((0x40000) - 2*PAGESIZE)
+/* We hardwire these, since we can't make libc calls. */
+#define PAGESIZE 4096
+#define ROUNDADDRUP(addr, size) ((addr + size - 1) & ~(size - 1))
+// We use global variables here, since both remap_vdso_and_vvar_regions and
+// unmap_memory_areas_and_restore_vdso use these variables. They're used to
+// track the temporary regions the vvar and vdso segments are remapped to, so
+// they are not unmapped later.
+uint64_t vvarStartTmp = 0;
+uint64_t vdsoStartTmp = 0;
```
jungan, line 507: Add `static` if they will only be used in this file.

configure-mana
fyshhh: yes we do. i've tested this version of `configure-mana` on Cori and it works.

contrib/mpi-proxy-split/split_process.cpp
fyshhh: this was used in an earlier version of the code, so we don't need to now. i will remove it.

include/procmapsarea.h
fyshhh: if this is removed, it causes an assertion error in `mmap.c/getNextAddr`.

src/mtcp/mtcp_restart.c
fyshhh: adding `static` strangely causes the `vdso/vvar order was different during ckpt` error.

PR #44, opened 2022-01-04
Merge feature/centos into master branch

fyshhh:
This pull request comprises a few commits made to the feature/centos branch to support building, launching, checkpointing and restarting MANA on CentOS, as well as some code generalizations. These commits have been tested on Cori and affirmed to work, by comparing the results from contrib/mpi-proxy-split/test with and without these commits.

This pull request is a follow-up to #40, but with a different source branch so changes can be made more easily.


jungan (approved):

gc00 (approved):
@fyshhh , I'm also approving the PR, so as not to delay this.
But I'd like to fix one thing before you push it in:

Right now, there is a nice file:
    &nbsp;&nbsp; doc/centos-tutorial.txt
But we don't point to it from anywhere else.  So, a new user has no way to discover this.
I suggest:
1.  Change the name to `doc/mana-centos7-tutorial.txt`  (Note also _centos7_ in the finlename.)
2.  Let's add to the top of both `contrib/mpi-proxy-split/README` and `contrib/mpi-proxy-split/INSTALL` a sentence, something like:  See doc/mana-centos7-tutorial.txt for instructions for installing on CENTOS 7.    [ and then add something like "====" on a line just below it. ]
3.   After merging this, please start a new PR in which you clean up some of the obsolete stuff in README and INSTALL.  I can help you with that PR.

I think the reviews from @jungan and me should be fine.  We don't need @rohgarg to review this.

PR #45, opened 2022-01-05
[MANA-37] Add version tracking to MANA

fyshhh:
This pull request adds version tracking to MANA, organized similarly to how criu does.

Note that this pull request will be merged into `master` - the base repository is currently set as `feature/centos` so it's easier to read, but will be changed once the commits in `feature/centos` is merged into `master`.


jungan (commented):
Did you add the git commit hash id?

We probably also want to add a cmdline option to those binaries, so something like `dmtcp_coordinator --version` will print the version number.

BTW, you should submit PR to master, this change is not related to CentOS, rebase your local branch on top of master, and then submit PR to master.

JainTwinkle:
Is the intention to keep track of both DMTCP's version and MANA's version? If so then should we rename the header file as mana-version.h? because the root directory of MANA has a `VERSION` file that points to DMTCP's version. Otherwise, MANA's `README.md` should be updated stating how to get either version's information.

fyshhh:
> Is the intention to keep track of both DMTCP's version and MANA's version? If so then should we rename the header file as mana-version.h? because the root directory of MANA has a `VERSION` file that points to DMTCP's version. In addition , MANA's `README.md` should be updated stating how to get either version's information.

yes, i think we are starting to track MANA's version, the same way criu tracks theirs. i think it would be ideal to merge both version files, just to have one centralized one that keeps track of all the versions of relevant packages. while we discuss whether this is viable, renaming it to `mana-version.h` is a good stop-gap.

fyshhh:
test this please

jungan:
> > Is the intention to keep track of both DMTCP's version and MANA's version? If so then should we rename the header file as mana-version.h? because the root directory of MANA has a `VERSION` file that points to DMTCP's version. In addition , MANA's `README.md` should be updated stating how to get either version's information.
> 
> yes, i think we are starting to track MANA's version, the same way criu tracks theirs. i think it would be ideal to merge both version files, just to have one centralized one that keeps track of all the versions of relevant packages. while we discuss whether this is viable, renaming it to `mana-version.h` is a good stop-gap.

I don't think we want to merge both versions, to my understanding, MANA is built on top of DMTCP, but they could dev&release separately. MANA should have its own version. (Each MANA version might be on top of a different DMTCP version though)

I also agree with @JainTwinkle that we should update the READ.me file on how to get the MANA version. We should add command-line option to `mana_coordinater` and/or other binary tools.

gc00:
I agree with the points by @jungan.
1.  MANA is built on top of DMTCP, but they could dev&release separately. MANA should have its own version. (Each MANA version might be on top of a different DMTCP version though)
2. I also agree with @JainTwinkle that we should update the READ.me file on how to get the MANA version. We should add command-line option to mana_coordinater and/or other binary tools. _(Now that we have a well-developed `manpages/mana.latex2man` file, perhaps you could add getting the MANA version to that file.  `mana.1` is built from `mana.latex2man`.)_

Since @jungan has been tracking this PR, I'll let @jungan take the lead on reviewing it.

PR #51, opened 2022-01-21
Fix compiler warnings

fyshhh:
This pull request fixes some compiler warnings generated when building MANA on CentOS - some were caused by some of my older PRs.

### DIFF
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -468,6 +468,8 @@ int discover_union_ckpt_images(char *argv[],
   return rank;
 }
 
+// This is to satisfy compiler warnings.
+#if 0
 NO_OPTIMIZE
 static unsigned long int
 mygetauxval(char **evp, unsigned long int type)
@@ -480,6 +482,7 @@ mygetauxval(char **evp, unsigned long int type)
   }
   return 0;
 }
+#endif
 
 NO_OPTIMIZE
 static int
@@ -512,7 +515,7 @@ uint64_t vdsoStartTmp = 0;
 static void
 remap_vdso_and_vvar_regions() {
   Area area;
-  int rc = 0;
+  void *rc = 0;
   uint64_t vvarStart = 0;
   uint64_t vdsoStart = 0;
   uint64_t vvarSize = 0;
@@ -527,11 +530,11 @@ remap_vdso_and_vvar_regions() {
 
   while (mtcp_readmapsline(mapsfd, &area)) {
     if (mtcp_strcmp(area.name, "[vvar]") == 0) {
-      vvarStart = area.addr;
-      vvarSize = area.size;
+      vvarStart = (uint64_t) area.addr;
+      vvarSize = (uint64_t) area.size;
     } else if (mtcp_strcmp(area.name, "[vdso]") == 0) {
-      vdsoStart = area.addr;
-      vdsoSize = area.size;
+      vdsoStart = (uint64_t) area.addr;
+      vdsoSize = (uint64_t) area.size;
     }
 
     if (vvarStart > 0 && vdsoStart > 0) {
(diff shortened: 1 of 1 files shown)


configure
```diff
@@ -4989,6 +4989,10 @@ if test -n "$ac_unrecognized_opts"; then
   esac
 fi
 
+CFLAGS="$CFLAGS -Werror"
+CPPFLAGS="$CPPFLAGS -Werror"
+CXXFLAGS="$CXXFLAGS -Werror"
```
fyshhh, line 4994: i'd preferably like to make this possible to disable (like through a `--disable-werror` flag), but i'm not sure how to exclude this flag from `$ac_unrecognized_opts`.

fyshhh:
i see these lines a few times in `./configure`, each testing different things:

```
if test "$use_logging" = "yes"; then
    $as_echo "#define LOGGING 1" >>confdefs.h
...
```

what's `confdefs.h` referring to here? i ran a search through the MANA directory and i couldn't find any file by that name.

contrib/mpi-proxy-split/split_process.cpp
```diff
@@ -333,7 +333,8 @@ startProxy()
       // Read from stdout of lh_proxy full lh_info struct, including orig memRange.
       close(pipefd_out[1]); // close write end of pipe
       // FIXME: should be a readall. Check for return error code.
-      if (read(pipefd_out[0], &lh_info, sizeof lh_info) < sizeof lh_info) {
+      if ((unsigned long) read(pipefd_out[0], &lh_info, sizeof lh_info)
+        < sizeof lh_info) {
```
jungan, line 337: `read` can return `-1`, make it `unsigned long` will be 2^32 - 1 and passed this condition check, which is not correct.

contrib/mpi-proxy-split/two-phase-algo.cpp
```diff
@@ -130,7 +130,8 @@ TwoPhaseAlgo::commit_begin(MPI_Comm comm)
   // Set state again incase we returned from beforeTrivialBarrier
   MPI_Request request;
   MPI_Comm realComm = VIRTUAL_TO_REAL_COMM(comm);
-  int flag = 0;
+  // Consider removing the #if 0 segment below?
```
jungan, line 133: I don't know what's the purpose of the `#if 0` bloc below, may keep it for debugging purposes.

But in C++, we usually define the variable right before we use it, so you can move this line to 145 if we don't want to remove that `#if 0` block.

src/mtcp/mtcp_restart.c
```diff
@@ -773,10 +775,12 @@ main(int argc, char *argv[], char **environ)
...
       Area heap_area;
       MTCP_ASSERT(getMappedArea(&heap_area, "[heap]") == 1);
-      start1 = max(heap_area.endAddr, lh_info.memRange.end);
+      start1 = (char *) max((uint64_t) heap_area.endAddr,
+                            (uint64_t) lh_info.memRange.end);
       Area stack_area;
       MTCP_ASSERT(getMappedArea(&stack_area, "[stack]") == 1);
-      end1 = min(stack_area.endAddr - 4 * GB, highMemStart - 4 * GB);
+      end1 = (char *) min((uint64_t) stack_area.endAddr - 4 * GB,
+                          (uint64_t) highMemStart - 4 * GB);
```
jungan, line 783: We should be very careful when performing `minus` on `uint64_t`, making sure the first is greater than the second.

src/mtcp/mtcp_restart.c
```diff
@@ -710,8 +714,6 @@ main(int argc, char *argv[], char **environ)
     // If we want to test this, we can add code to do a trial mremap with a page
     //   before vvar and after vdso, and verify that we get an EFAULT.
     // In May, 2020, on Cori and elsewhere, vvar is 3 pages and vdso is 2 pages.
-    char *vdsoStart = (char *)mygetauxval(environ, AT_SYSINFO_EHDR);
```
jungan, line 713: What is this for and why removing this line?

jungan (changes requested):

src/mtcp/mtcp_restart.c
fyshhh: this is a currently unused variable that was previously used when determining where to temporarily map the `vdso`/`vvar` segments to. there used to be a function `mysetauxval` that set `AT_SYSINFO_EHDR` so this line could read for it, but since that function is no longer in use we can remove this one as well.

PR #52, opened 2022-01-26
[MANA-76] Fix mana_restart bug when vvar/vdso segment does not exist

fyshhh:
This pull request fixes a bug occurring in `mana_restart` when a `vdso` or `vvar` segment doesn't exist. A previous pull request updated the function to dynamically find the sizes of these segments to map to, which resulted in undefined behaviour in some functions since the sizes were previously hardcoded.

### DIFF
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -1453,8 +1453,13 @@ unmap_memory_areas_and_restore_vdso(RestoreInfo *rinfo, LowerHalfInfo_t *lh_info
   }
   mtcp_sys_close(mapsfd);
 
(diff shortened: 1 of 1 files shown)


gc00 (approved):
LGTM.  Thanks for catching this.  It's likely that I missed this case when i submitted an earlier RP to extend the capability of mtcp_restart.

Could you also submit this PR to https://github.com/dmtcp/dmtcp (master branch)?
Thanks.

fyshhh:
@gc00 i've opened a PR [here](https://github.com/dmtcp/dmtcp/pull/926), but i can't add you as a reviewer.

PR #58, opened 2022-02-16
Add more function pointer types in record-replay

xuyao0127:
Add MPI_Comm_copy_attr_function* and MPI_Comm_delete_attr_function* in FncArg class for record-replay.

### DIFF
--- a/contrib/mpi-proxy-split/record-replay.h
+++ b/contrib/mpi-proxy-split/record-replay.h
@@ -93,6 +93,8 @@ namespace dmtcp_mpi
     TYPE_VOID_PTR,
     TYPE_VOID_CONST_PTR,
     TYPE_MPI_USER_FNC,
+    TYPE_COMM_COPY_FNC,
+    TYPE_COMM_DELETE_FNC,
   };
 
   // Struct for saving arbitrary function arguments
@@ -184,6 +186,16 @@ namespace dmtcp_mpi
     {
       return *(MPI_User_function**)_data;
     }
+
+    operator MPI_Comm_copy_attr_function*() const
+    {
+      return *(MPI_Comm_copy_attr_function**)_data;
+    }
+
+    operator MPI_Comm_delete_attr_function*() const
+    {
+      return *(MPI_Comm_delete_attr_function**)_data;
+    }
   };
 
   template<typename T>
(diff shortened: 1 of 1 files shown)


contrib/mpi-proxy-split/record-replay.h
```diff
@@ -184,6 +186,16 @@ namespace dmtcp_mpi
...
     }
+
+    operator MPI_Comm_copy_attr_function*() const
+    {
+      return *(MPI_Comm_copy_attr_function**)_data;
+    }
+
+    operator MPI_Comm_delete_attr_function*() const
+    {
+      return *(MPI_Comm_delete_attr_function**)_data;
```
dahongli, line 197: These two function pointers are used in MPI_Comm_create_keyval. Could you add a test case for MPI_Comm_create_keyval to test them?

dahongli (commented):
The change looks good. Please add a test case for these new types. Thanks Yao!

gc00 (commented):
@xuyao0127, I think that we don't want to add anything in the `record-replay.cpp` file for the `copy_attr_function` and `delete_attr_function`.  @fyshhh is implementing this functionality for MANA right now.

Our conclusion is that these attribute functions (and attribute_val) are declared independently for each MPI process.  There is no communication between MPI ranks in this case.

Because of this, @fyshhh intends to implement communicator attributes locally, using STL ordered mappings.  We will still want wrappers for the `MPI_Comm_create_keyval` and related calls.  But these wrappers will _not_ call the lower half.  They will simply set something in an ordered mapping that is found via a global variable.

Finally, the callback functions `copy_attr_function` and `delete_attr_function` will later be called from inside the MANA wrappers for `MPI_Comm_free`, `MPI_Comm_dup`, etc.  These attribute functions will be called _immediately_ and outside of the `DISABLE/ENABLE_CKPT` macros.  It's fine for us to checkpoint in the middle of this attribute operation.

We'll have to look at use cases to see if some application code is using this ability in order to create a temporary file associated with the attribute, and later delete it.  But even in that case, DMTCP knows how to save and restore other files.

If you still have questions, let's discuss this interactively.

xuyao0127:
I agree the new approach is better.

PR #60, opened 2022-02-22
Fix c9d951015 for 'core' for gdb-dmtcp-utils

gc00:
The last enhancement of `util/gdb-dmtcp-utils` had a regression, in which the ability to use this i GDB with core files (using `info proc mappings`) was lost.  This PR fixes that regression, and polishes the Python code a little more.


fyshhh (approved):
lgtm!

PR #61, opened 2022-03-04
Minor fix to commit 84d37fb01 for gdb-dmtcp-utlls

gc00:
This was a weird bug, and I still don't know how to reproduce it deterministically.  But @xuyao0127 encountered this bug while working in MANA, doing something like this patter:
```
add-symbol-files-all
bt
frame 4
whereis-address 0xffff7ade0000
```

It appears to be a weird interaction between GDB-8 and its Python interface.  When the bug occurs, `whereis-address` thinks there is a `"` in the address.  At that point, `(gdb) print "abc"` fails with the same error, but `(gdb) print 'abc'` succeeds (even though it should not succeed, since `'` is used to indicate the beginning of the full signature for a C++ method or variable.


xuyao0127 (approved):
LGTM.

PR #62, opened 2022-03-06
Collective p2p

gc00:
This is a first version of converting collective calls to point-to-point.  The goal is _not_ a production version of MANA.  Rather, the goal is a lower performing version of MANA that does not make any collective calls.  This makes it easy to debug any bugs related to collective calls.  We can see if the alternative implementation in terms of point-to-point calls has the same bug.  And we can selectively define only some of the collective calls as point-to-point.

As stated in one of the commits:
```
     * Include the file mpi_collective_p2p.c in
          mpi_collective_wrappers.cpp, instead of original wrappers
     * The collective communication functions in mpi_collective_p2p.c
          make calls solely to point-to-point calls in the upper half,
          which in turn then call the point-to-point functions in the lower half.
```

This adds `mpi-wrappers/mpi_collective_p2p.c` and defines `COLLECTIVE_P2P` in `mpi-wrappers/mpi_collective_wrappers.cpp`.
This causes `mpi_collective_wrappers.cpp` to include `mpi_collective_p2p.c`, and skip defining the wrappers that would call the collective function in the lower half.

This defines almost all of the collective communication functions defined by MANA.  This still needs to be tested on some larger programs, like VASP.  (It doesn't yet define `MPI_Ispan`, which appears to be used by VASP.  But we can still allow VASP to call the upper-half stub function for `MPI_Ispan`, which will call the lower half.  I'll add a definition for `MPI_Ispan` later.)

### DIFF
--- a/contrib/mpi-proxy-split/mpi-wrappers/mpi_collective_p2p.c
+++ b/contrib/mpi-proxy-split/mpi-wrappers/mpi_collective_p2p.c
@@ -1,0 +1,627 @@
+#include <assert.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <mpi.h>
+
+// NOTE:  Do 'google MPI standard 3.1' to see the spec in the standard.
+//        This first version does not consider intracommunicators.
+//        This version does not test for return values.
+
+#define PROLOG_rank_size \
+  int rank; \
+  int size; \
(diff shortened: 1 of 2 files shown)


gc00:
I meant to add this PR to the `feature/collectiveToP2p`, not to `master` branch.  I'm closing this PR in favor of PR #63.

PR #63, opened 2022-03-06
Collective p2p

gc00:
This is a first version of converting collective calls to point-to-point.  The goal is _not_ a production version of MANA.  Rather, the goal is a lower performing version of MANA that does not make any collective calls.  This makes it easy to debug any bugs related to collective calls.  We can see if the alternative implementation in terms of point-to-point calls has the same bug.  And we can selectively define only some of the collective calls as point-to-point.

As stated in one of the commits:
```
     * Include the file mpi_collective_p2p.c in
          mpi_collective_wrappers.cpp, instead of original wrappers
     * The collective communication functions in mpi_collective_p2p.c
          make calls solely to point-to-point calls in the upper half,
          which in turn then call the point-to-point functions in the lower half.
```

This adds `mpi-wrappers/mpi_collective_p2p.c` and defines `COLLECTIVE_P2P` in `mpi-wrappers/mpi_collective_wrappers.cpp`.
This causes `mpi_collective_wrappers.cpp` to include `mpi_collective_p2p.c`, and skip defining the wrappers that would call the collective function in the lower half.

This defines almost all of the collective communication functions defined by MANA.  This still needs to be tested on some larger programs, like VASP.

(It doesn't yet define `MPI_Ispan`, which appears to be used by VASP.  But we can still allow VASP to call the upper-half stub function for `MPI_Ispan`, which will call the lower half.  I'll add a definition for `MPI_Ispan` later.)

**FIXED NOW:**   This code uses `MPI_Reduce_local`, and MANA originally didn't have a wrapper for `MPI_Reduce_local`.  This PR now includes a commit to implement `MPI_Reduce_local`.

### DIFF
--- a/contrib/mpi-proxy-split/mpi-wrappers/mpi_collective_p2p.c
+++ b/contrib/mpi-proxy-split/mpi-wrappers/mpi_collective_p2p.c
@@ -432,54 +432,6 @@ int MPI_Reduce(const void* sendbuf, void* recvbuf, int count,
   }
   return MPI_SUCCESS;
 }
-#else
-// FIXME:  This wuld still need to implement MPI_Reduce over two processes
-int MPI_Reduce(const void* sendbuf, void* recvbuf, int count,
-               MPI_Datatype datatype, MPI_Op op, int root, MPI_Comm comm) {
-  PROLOG_rank_size;
-  MPI_Aint lower_bound;
-  MPI_Aint extent;
-  MPI_Type_get_extent(datatype, &lower_bound, &extent);
-  int inplace = (sendbuf == MPI_IN_PLACE);
-  if (inplace && rank == root) {
-    memcpy(recvbuf, sendbuf, count * extent);
-  }
-  // It would have been nice if MPI had a way to use 'op' locally on array.
-  int other_root = (root != 0 ? 0 : 1);
-  char *tmp_recvbuf = NULL;
-  if (rank == other_root) {
-    tmp_recvbuf = malloc(count * extent);
-    MPI_Gather(sendbuf, count, datatype, tmp_recvbuf, count, datatype,
-               other_root, comm);
-  } else {
-    MPI_Gather(sendbuf, count, datatype, NULL, 0, datatype,
-               other_root, comm);
-  }
-  if (rank == root || rank == other_root) {
-    MPI_Group group, tmpgroup;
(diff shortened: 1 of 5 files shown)


contrib/mpi-proxy-split/mpi-wrappers/mpi_collective_p2p.c
```diff
@@ -0,0 +1,630 @@
...
+  }
+
+  if (rank == root) {
+    // Gather data into tmp_sendbuf at root
+    int i;
+    char *tmp_sendbuf = (char *)malloc(count * extent);
+    MPI_Gather(sendbuf, count, datatype, tmp_sendbuf, count, datatype,
+               root, comm);
+    // Initialize tmp_recvbuf from sendbuf at rank 0
+    char *tmp_recvbuf = malloc(count * extent);
```
jungan, line 415: Unfortunately, g++ doesn't allow implicit conversion.
gcc allows it, but this file was included in cpp so will be compiled by g++

jungan (commented):
Hi @gc00 I think it will take a long time for me to check the correctness of your code with MPI spec. What's your suggestion the next step for me to do?
I think we probably want to test this with the test program first. I will try it tomorrow.

gc00:
Thanks.  Yes, let's run this on some test programs -- both the MANA test suite and VASP.

contrib/mpi-proxy-split/mpi-wrappers/mpi_collective_p2p.c
```diff
@@ -0,0 +1,638 @@
...
+    for (i = 0; i < size; i++) {
+      if (i != root) {
+        MPI_Send(sendbuf + i*sendcount*sendextent, sendcount, sendtype,
+                 i, 0, comm);
+      }
+    }
+  } else { // else: rank != root
+    for (i = 0; i < size; i++) {
+      if (i != root) {
+        MPI_Recv(recvbuf, recvcount, recvtype, i, 0, comm, MPI_STATUS_IGNORE);
```
dahongli, line 214: The source should be "root". Replace "I" with "root".

gc00: @dahongli , Great catch!  Thanks.  I've pushed in a fixed version now.

PR #66, opened 2022-03-09
bug fixes for collective to p2p

xuyao0127:
It seems checkpointing is working fine with these fixes. However, the feature/dmtcp-master branch still can't restart because of another bug. So I am not sure if we need other fixes.

I decided to move the MPI_Alltoall_internal into p2p_drain_send_recv.cpp. The function is only used in this file, and it's not really a wrapper. The purpose of this function is letting us use it internally without applying the two-phase-commit algorithm.

### DIFF
--- a/contrib/mpi-proxy-split/mpi-wrappers/mpi_collective_p2p.c
+++ b/contrib/mpi-proxy-split/mpi-wrappers/mpi_collective_p2p.c
@@ -367,6 +367,7 @@ int MPI_Alltoallv(const void* sendbuf, const int sendcounts[],
   return MPI_SUCCESS;
 }
 
+#ifdef ADD_UNDEFINED
 int MPI_Alltoallw(const void* sendbuf, const int sendcounts[],
                   const int sdispls[], const MPI_Datatype sendtypes[],
                   void* recvbuf, const int recvcounts[], const int rdispls[],
@@ -375,6 +376,7 @@ int MPI_Alltoallw(const void* sendbuf, const int sendcounts[],
   ABORT();
   return -1;
 }
+#endif
 
 // MPI standard 3.1:  Section 5.9
 /* NOTE:  MPI-3.1 standard (Section 5.9.1):
@@ -440,6 +442,7 @@ int MPI_Allreduce(const void* sendbuf, void* recvbuf, int count,
   return MPI_SUCCESS;
 }
 
+#ifdef ADD_UNDEFINED
 // MPI standard 3.1:  Section 5.10
 int MPI_Reduce_scatter_block(const void* sendbuf, void* recvbuf,
                              int recvcount, MPI_Datatype datatype, MPI_Op op,
@@ -448,6 +451,8 @@ int MPI_Reduce_scatter_block(const void* sendbuf, void* recvbuf,
   ABORT();
   return -1;
 }
+#endif
+
 int MPI_Reduce_scatter(const void* sendbuf, void* recvbuf,
                        const int recvcounts[], MPI_Datatype datatype, MPI_Op op,
                        MPI_Comm comm) {
@@ -496,7 +501,7 @@ int MPI_Scan(const void* sendbuf, void* recvbuf, int count,
   }
   return MPI_SUCCESS;
 }
-#if ADD_UNDEFINED
+#ifdef ADD_UNDEFINED
 int MPI_Exscan(const void* sendbuf, void* recvbuf, int count,
              MPI_Datatype datatype, MPI_Op op, MPI_Comm comm) {
   fprintf(stderr, "%s not implemented\n", __FUNCTION__);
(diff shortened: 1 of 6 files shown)


contrib/mpi-proxy-split/mpi-wrappers/mpi_collective_wrappers.cpp
```diff
@@ -273,24 +273,6 @@ USER_DEFINED_WRAPPER(int, Reduce_scatter,
...
-  MPI_Datatype realSendType = VIRTUAL_TO_REAL_TYPE(sendtype);
-  MPI_Datatype realRecvType = VIRTUAL_TO_REAL_TYPE(recvtype);
-  JUMP_TO_LOWER_HALF(lh_info.fsaddr);
-  retval = NEXT_FUNC(Alltoall)(sendbuf, sendcount, realSendType, recvbuf,
-      recvcount, realRecvType, realComm);
-  RETURN_TO_UPPER_HALF();
-  DMTCP_PLUGIN_ENABLE_CKPT();
-  return retval;
-}
-
```
jungan, line 293: It will be called by `Alltoall`, so it won't compile when we want to use collective version.

BTW, are we plan to eventually merge this branch to master, and use a MACRO switch to control it?

xuyao0127:
@jungan I have undone the change about MPI_Alltoall_internal and added a copy of the function in the collective_p2p wrapper. Now it should work in both cases.

contrib/mpi-proxy-split/mpi-wrappers/mpi_collective_wrappers.cpp
gc00: Yes, as @jungan writes, "we plan to eventually merge this branch to master, and use a MACRO switch to control it."

The MACRO switch is `MPI_COLLECTIVE_P2P` in `mpi_collective_wrappers.cpp`.  When we merge this into master, we should add a comment in a Makefile that can define `MPI_COLLECTIVE_P2P`, but it should be undefined by default.

@xuyao0127 says that when testing on VASP, this doesn't yet restart.  I plan to work with him this afternoon to try to fix that.  Once we fix that, we can then do another PR.  And once that PR is accepted, we can then merge this into master (cherry-pick the commits from this feature branch, as a new PR against master).

contrib/mpi-proxy-split/mpi-wrappers/mpi_collective_p2p.c
```diff
@@ -413,7 +413,7 @@ int MPI_Reduce(const void* sendbuf, void* recvbuf, int count,
   if (rank == root) {
     // Gather data into tmp_sendbuf at root
     int i;
-    char *tmp_sendbuf = (char *)malloc(count * extent);
+    char *tmp_sendbuf = (char *)malloc(count * extent * size);
```
gc00, line 416: I believe this commit is incorrect.   `size` comes from `MPI_Comm_size` in `PROLOG_rank_size`.  The original `extent` should be correct.  I propose to remove this commit.

Would it be better if we rename `PROLOG_rank_size` to `PROLOG_Comm_rank_size`?

contrib/mpi-proxy-split/mpi-wrappers/mpi_collective_p2p.c
```diff
@@ -643,6 +643,24 @@ int MPI_Iexscan(const void* sendbuf, void* recvbuf, int count,
 }
 #endif  // #ifdef ADD_UNDEFINED
 
+int
+MPI_Alltoall_internal(const void *sendbuf, int sendcount,
```
gc00, line 647: This commit is good for now.

But I'd like to do a follow-on PR that modifies the code here for MPI_Alltoall_internal so that it makes calls to MPI_Send/Recv in the upper half, before calling the lower half.  We don't want it to directly make a call to MPI_Alltoall in the lower half.  Our principle should be to never call the lower half with a _collective communication_ call.  Then if there's a bug, we can guarantee that it's about the point-to-point layer.

gc00 (changes requested):
@xuyao0127, could you please remove the commit `[Fixed a bug: MPI_Reduce's tmp_sendbuf wrong size]`? 
 As indicated in the inline comment, I believe that this commit actually creates a bug in code that was already correct.

Could you then re-test this, to see if it still works, and there's no regression?  Thanks.

Also, in a future commit, I will change `PROLOG_rank_size` to `PROLOG_Comm_rank_size` to avoid future confusion.

contrib/mpi-proxy-split/mpi-wrappers/mpi_collective_p2p.c
gc00: Thanks for the correction.  It looks good.

contrib/mpi-proxy-split/mpi-wrappers/mpi_collective_p2p.c
gc00: Thanks for the comment.

gc00 (approved):
LGTM.

PR #67, opened 2022-03-09
Add several Fortran90 test samples

jungan:
Several test cases download from https://people.math.sc.edu/Burkardt/f_src/mpi/mpi.html

Managed to make `wave_mpi` and `poisson_nonblock_mpi` to run a much longer time by changing some parameters.

Not user how to make `quad_mpi` `day1_mp` to run longer, I don't want to use `sleep` which I think will hide potential issues, how about simply add more loops to do the same thing again and again?

### DIFF
--- a/contrib/mpi-proxy-split/test/Makefile
+++ b/contrib/mpi-proxy-split/test/Makefile
@@ -33,7 +33,8 @@ FILES=mpi_hello_world \
       Abort_test Allreduce_test Alltoall_test Alltoallv_test \
       Allgather_test Group_size_rank Type_commit_contiguous \
       Irecv_test Alloc_mem \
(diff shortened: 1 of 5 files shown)


gc00 (approved):
@jungan , Thanks for these test files.  I'll approve this in advance, but in your commit message, `Add several Frotran90 test samples`, could you fix the typo (Frotran->Fortran)?

Otherwise, it's good to go.

Could you also (outside of this PR) look into a good way to run these tests with Jenkins (for the master branch)?  We don't need to regularly run the other simple codes i the test directory, but these new codes are similar to what we see in applications, and they can provide good, partial coverage for MPI.  Thanks.

PR #70, opened 2022-03-10
Log value instead of pointer in MPI_Cart_shift

jungan:
When testing `poisson_nonblock_mpi`, we saw following error when restart:
```
[40000] ERROR at record-replay.cpp:649 in restoreCartShift; REASON='JASSERT(oldsrc == rank_source && olddest == rank_dest) failed'
     oldsrc = -42144
     olddest = -42140
     rank_source = -1
     rank_dest = 1
Message: Different ranks
poisson_nonblock_mpi.mana.exe (40000): Terminating...
```

It turns out it's because MANA records the pointer instead of the real value.

BTW, I also saw this in the code:

```
  if (retval == MPI_SUCCESS) {
    // FIXME: Virtualize rank?
    int oldsrc = rec.args(3);
    int olddest = rec.args(4);
    JASSERT(oldsrc == rank_source && olddest == rank_dest)
           (oldsrc)(olddest)(rank_source)(rank_dest).Text("Different ranks");
  }
```

Looked at the `FIXME`, wondering is it possible the ranks change after C/R? Even we do, will this be a problem?

### DIFF
--- a/contrib/mpi-proxy-split/mpi-wrappers/mpi_cart_wrappers.cpp
+++ b/contrib/mpi-proxy-split/mpi-wrappers/mpi_cart_wrappers.cpp
@@ -129,7 +129,7 @@ USER_DEFINED_WRAPPER(int, Cart_shift, (MPI_Comm) comm, (int) direction,
   RETURN_TO_UPPER_HALF();
   if (retval == MPI_SUCCESS && MPI_LOGGING()) {
     LOG_CALL(restoreCarts, Cart_shift, comm, direction,
-             disp, rank_source, rank_dest);
+             disp, *rank_source, *rank_dest);
   }
   DMTCP_PLUGIN_ENABLE_CKPT();
   return retval;


gc00 (approved):
LGTM.

@jungan, in answer to your question about the FIXME, right now, MANA has a performance problem (and/or a correctness problem) for Cartesian coordinates.  When we restart, we replay the MPI_Cart_create() command.  That will create the Cartesian coordinates during restart, based on an efficient mapping from the MPI Cartesian coordinates and the underlying hardware interconnect in the network.  But there is a correctness problem:  A process might have had rank r1 and Cartesian coordinate (i1, j1) before restart, and then r1 and (i2, j2) after restart.  To avoid this inconsistency, we could virtualize the rank to vr1, and then have VRTUAL_TO_REAL translate vr1 to r2.

But right now, we are thinking of a different design that should bring similar benefits.

PR #71, opened 2022-03-10
Test env var DEFINE_mpi_collective_p2p & re-build

gc00:
This commit supports the `DEFINE_mpi_collective_p2p` environment variable.  When it is set, and if re-building any files in   `contrib/mpi-proxy-split/mpi-wrapper`, then it will include `contrib/mpi-proxy-split/mpi-wrapper/mpi_collective_p2p.c` when building.  At runtime, this will translate MPI collective communication calls to MPI_Send/Recv. 

Please test this PR on Cori, when you review it.  I haven't done enough testing of the various cases.


dahongli (approved):
Looks good. Thanks Gene!

PR #73, opened 2022-03-11
[MANA-77] Fix ip command not found bug in configure-mana

fyshhh:
This pull request fixes a bug where the `ip` command cannot be found in the default path for root users by adding `/usr/sbin` to the path temporarily while executing the script.

### DIFF
--- a/configure-mana
+++ b/configure-mana
@@ -1,13 +1,13 @@
 #!/bin/sh
 
 # These optional autoconf variables will be configured
-#     via contrib/mpi-proxy-split/Makefie_config.in
+#     via contrib/mpi-proxy-split/Makefile_config.in
 
 # Do not add spaces to the next two lines.
 # If MPI will be used on a different computer than where you configure,
 #   then maybe replace this with the result of 'ip addr' on the compute nodes.
 MPI_ETHERNET_INTERFACE=\
-`ip addr |grep -B1 link/ether | head -1 |sed -e 's%[^ ]*: \([^ ]*\): .*%\1%'`
+`PATH=$PATH:/usr/sbin ip addr |grep -B1 link/ether | head -1 |sed -e 's%[^ ]*: \([^ ]*\): .*%\1%'`
 
 # Note: modify this path according to your environment.
 if [[ -z "${MPI__INSTALL_DIR}" ]]; then


configure-mana
```diff
@@ -1,7 +1,10 @@
 #!/bin/sh
 
 # These optional autoconf variables will be configured
-#     via contrib/mpi-proxy-split/Makefie_config.in
+#     via contrib/mpi-proxy-split/Makefile_config.in
+
+# Temporarily add /usr/sbin to the PATH in case ip can't be found.
+export PATH=$PATH:/usr/sbin
```
jungan, line 7: We should be very careful on changing the global environment variables, if `ip` is the only one that has the problem why not just change that to `/usr/sbin/ip` ?

Another question, why `PATH` doesn't have `/usr/sbin` here? I think this should be there by default

jungan: 1. If `ip` is not in `/usr/sbin`, how your fix can fix the problem?
2. I don't know where you found that, it doesn't make sense to me is that adding `/usr/sbin` doesn't grant you any permission. How can that be a problem?

fyshhh: 1. If it's not in `PATH` or `/usr/sbin` then it means that the user doesn't have it and should install it.
2. I did a bit more digging and it depends on the OS; for instance Debian doesn't have it by default whereas CentOS does.

gc00: Sorry that it took me so long to review this.

Here's what I suggest, @fyshhh:
1.  Delete the current changes that you have in configure-mana
2. In `configure-mana`, change, as below.
FROM:
```
MPI_ETHERNET_INTERFACE=\
`ip addr |grep -B1 link/ether | head -1 |sed -e 's%[^ ]*: \([^ ]*\): .*%\1%'`
```
TO:
```
MPI_ETHERNET_INTERFACE=\
`PATH=$PATH:/usr/sbin ip addr |grep -B1 link/ether | head -1 |sed -e 's%[^ ]*: \([^ ]*\): .*%\1%'`
```

This seems to accomplish your intended goal, while also not creating the danger mentioned by @jungan of adding `/usr/sbin` in the path of other commands.

When you've done this, please tag @jungan and me, so that we can give this a quick review.  Thanks.

gc00 (changes requested):
Please see the earlier comment about using
```
MPI_ETHERNET_INTERFACE=\
`PATH=$PATH:/usr/sbin ip addr |grep -B1 link/ether | head -1 |sed -e 's%[^ ]*: \([^ ]*\): .*%\1%'`
```
to accomplish your goal.

fyshhh:
@gc00, @jungan this PR is ready for review and merging.

jungan (approved):

gc00 (approved):
LGTM.

PR #81, opened 2022-03-17
Fix python not being found on CentOS 8 Stream

fyshhh:
This pull request updates the instances of `python` in MANA to `python3` instead, motivated by the `python` command not being found on CentOS 8 Stream (it doesn't have a default `python`; instead the version has to be specified). This also adds Python 3 installation to the CentOS 7 tutorial.

### DIFF
--- a/contrib/mpi-proxy-split/mpi-wrappers/Makefile
+++ b/contrib/mpi-proxy-split/mpi-wrappers/Makefile
@@ -32,12 +32,6 @@ override CXXFLAGS += -g3 -O0 -fPIC -I${DMTCP_INCLUDE} -I${JALIB_INCLUDE} \
                      -I${PLUGIN_HALF_SRCDIR} -I${DMTCP_ROOT}/src \
                      -I${LOWER_HALF_SRCDIR} -std=c++11
 
-ifeq ($(shell which python),python)
-  PYTHON = python
-else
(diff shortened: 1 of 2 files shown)


contrib/mpi-proxy-split/mpi-wrappers/Makefile
```diff
@@ -32,6 +32,12 @@ override CXXFLAGS += -g3 -O0 -fPIC -I${DMTCP_INCLUDE} -I${JALIB_INCLUDE} \
                      -I${PLUGIN_HALF_SRCDIR} -I${DMTCP_ROOT}/src \
                      -I${LOWER_HALF_SRCDIR} -std=c++11
 
+ifeq ($(shell which python),python)
+  PYTHON = python
+else
+  PYTHON = python2.7
```
jungan, line 38: Have you tried `python3`?
I think python2 is depreciated, we should think about upgrading the script if we have problems with python3.

And here, if we cannot work with python3, then we will have trouble if `python` is linked to `python3`

fyshhh: `python3` is indeed working. i can change the script to use that instead and remove the python part from the tutorial entirely.

jungan (approved):

PR #82, opened 2022-03-18
Set fs in user space

gc00:
This is based on the branch of @JainTwinkle.  Here, her commit is preserved as 82adb5e3bb.

This code should auto-detect if the fSGSBASE patch is available in the Linux kernel.  If so, set the FS register in user space and avoid the cost of a Linux kernel.

_I'm not sure why, but I'm always getting "illegal instruction" when compiling the code in configure.ac, even when the FSGSBASE patch is available._

_The program used to test for FSGSBASE is:_
```
int
main(void)
{
  // FIXME:  This always crashed, even with the FSGSBASE patch.
  unsigned long fsbase;
  asm volatile("rdfsbase %0" : "=r" (fsbase) :: "memory");
  asm volatile("wrfsbase %0" :: "r" (fsbase) : "memory");
  asm volatile("rdfsbase %0" : "=r" (fsbase) :: "memory");
  if (fsbase == 0) {
    return 0;
  } else {
    return 1;
  }
} 
```


gc00:
We are still waiting to hear from HPE/Cray on how to enable the FSGSBASE instructions on Cori.  Once we can test this on Cori, we can review this PR and hopefully push it in.

gc00:
This is no longer needed.  The equivalent code is now in [https://github.com/dmtcp/dmtcp/pull/1011] (PR 1011 of DMTCP).  when we update the DMTCP submodule, that will be inherited into MANA.

PR #72, opened 2022-03-11
Merge feature/collectiveToP2P to master

gc00:
This PR is for merging and rebasing origin:feature/collectiveToP2P on top of master.

In this PR, I've first created a branch, merge-collectiveToP2P, which has been rebased to include all the latest commits from master.  The PR is equivalent to simply adding origin:feature/collectiveToP2P as a PR for rebasing on top of origin, but this way, Jenkins can do a small test using the latest master branch+collectiveToP2P.

As noted in `manpages/mana.1` (or `manpages/mana.latex2man`), this should not produce any different behavior, _unless_ the environment variable `DEFINE_mpi_collective_p2p` is set before building MANA, or when re-building the MANA plugin.

Please do a quick test of building MANA in the origin branch with and without `DEFINE_mpi_collective_p2p`, to verify that both work as intended.

PR #83, opened 2022-03-19
Fix bug in commit 79bd117: affects collectiveToP2P

gc00:
This fixes commit 79bd117 used in PR #72 (actually from an earlier PR, but the commit was later captured in feature/collectiveToP2P and merged into master).

The earlier commit made MANA work when `MPI_COLLECTIVE_P2P` was defined, but it stopped working when `MPI_COLLECTIVE_P2P` was not defined.  (The bug was that the commit implicitly assumed that the C and C++ symbols for `MPI_Alltoall_internal` were the same.)

In this version, we _only_ define the C++ version of `MPI_Alltoall_internal`, and it is available in `mpi-wrappers/mpi_collective_wrappers.cpp:MPI_Alltoall_internal`, independently of whether `MPI_COLLECTIVE_P2P` is defined or not.

===

This also fixes a bug in the attempt of PR #80 (commit 838aa20) to fix the earlier bug.  PR #80 would cause `p2p_drain_send_recv.cpp` to use the C version of `MPI_Alltoall_internal` when `MPI_COLLECTIVE_P2P` is defined.  But the C version of that function was going to generate extra point-to-point calls at checkpoint time, when `p2p_drain_send_recv.cpp` tries to drain the point-to-point calls.  So, checkpoint would fail for any MPI application with pending point-to-point calls in this situation.
    Also, I was seeing a problem with and undefined C symbol for  `MPI_Alltoall_internal`  when `MPI_COLLECTIVE_P2P` was not defined, but I've lost track of why by now.

===

@xuyao0127, @fyshhh , and @jungan, please check my logic on this.  I'm trying to fix commits that created this logic error.  And as part of this, I'm trying to also ensure that there's only one copy of `MPI_Alltoall_internal`!  This will avoid the logic traps that we fell into with the two preceding commits.

### DIFF
--- a/contrib/mpi-proxy-split/mpi-wrappers/Makefile
+++ b/contrib/mpi-proxy-split/mpi-wrappers/Makefile
@@ -61,6 +61,8 @@ mpi_collective_wrappers.o: mpi_collective_wrappers.cpp ../virtual-ids.h \
                            mpi_collective_p2p.c
 	if [ ! -z "$$MPI_COLLECTIVE_P2P" ]; then \
 	  DEFINE_mpi_collective_p2p="-DMPI_COLLECTIVE_P2P"; \
+	else \
+	  DEFINE_mpi_collective_p2p=""; \
 	fi && \
 	${MPICXX} ${CXXFLAGS} -Wall -g3 -O0 -DDEBUG \
 	  $$DEFINE_mpi_collective_p2p -fPIC -I../../../include \
--- a/contrib/mpi-proxy-split/mpi-wrappers/mpi_collective_p2p.c
+++ b/contrib/mpi-proxy-split/mpi-wrappers/mpi_collective_p2p.c
@@ -662,11 +662,17 @@ int MPI_Iexscan(const void* sendbuf, void* recvbuf, int count,
 }
 #endif  // #ifdef ADD_UNDEFINED
 
+#if 0
+//FIXME:  Delete this, once we confirm it's not needed.  When MPI_COLLECTIVE_P2P is
+//        defined, the file p2p_drain_send_recv.cpp should call the C++ version
+//        of this, defined in mpi-wrappers/mpi_collective_wrappers.cpp.  (See the
+//        comment in the latter file.)  That is a call to the C++ version of
+//        MPI_Alltoall_internal.  So, this C version should never be used.
 // This routine is called from mpi-proxy-split/p2p_drain_send_recv.cpp.
 // The code in that file is invoked only at checkpoint time. It uses
 // MPI_Alltoall to drain any remaining Send/Recv messages.
 // So, it must not convert MPI_Alltoall_internal to use use point-to-point
-// communication. This is an excption that makes a collective call to the
+// communication. This is an exception that makes a collective call to the
 // lower half.
 int
 MPI_Alltoall_internal(const void *sendbuf, int sendcount,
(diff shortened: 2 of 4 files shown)


jungan:
> This also fixes a bug in the attempt of PR https://github.com/mpickpt/mana/pull/80 (commit https://github.com/mpickpt/mana/commit/838aa20902a59bf06c8d089171ec1c9bd265943f) to fix the earlier bug. PR https://github.com/mpickpt/mana/pull/80 would cause p2p_drain_send_recv.cpp to use the C version of MPI_Alltoall_internal when MPI_COLLECTIVE_P2P is defined. But the C version of that function was going to generate extra point-to-point calls at checkpoint time, when p2p_drain_send_recv.cpp tries to drain the point-to-point calls. So, checkpoint would fail for any MPI application with pending point-to-point calls in this situation.
Also, I was seeing a problem with and undefined C symbol for MPI_Alltoall_internal when MPI_COLLECTIVE_P2P was not defined, but I've lost track of why by now.

I don't quite understand what problem we have, actually, it's me who asked @fyshhh to fix the issue that "undefined C symbol for MPI_Alltoall_internal when MPI_COLLECTIVE_P2P was not defined", and that is what #80 was trying to do. I checked the C++ and C version of `MPI_Alltoall_internal `, they're exactly the same. I agree that your version is cleaner but why the C version will generate extra point-to-point calls? Isn't `MPI_Alltoall_internal` just jump to the lower half and call `MPI_Alltoall`?

jungan (approved):

gc00:
@jungan,  The isssue is that collectiveToP2P is intended to convert all collective calls to P2P during normal _runtime_.  Then, when it's time to checkpoint, MANA does:
1.  continue to execute until there are no collective calls in the lower half; and
2. then suspend all user threads, and use _only_ collective calls in an effort to drain any _user-generated_ pending point-to-point calls into buffers on the side of the receiver (the checkpoint thread should not be adding MANA-generated point-to-point calls that will be drained into buffers of user pending msg's), and then
3. do a checkpoint.

The problem with PR #80 is that after the user threads were suspended, it was converting MPI_Alltoall_internal to call the point-to-point versoin of MPI_Alltoall, which would then create MANA-generated point-to-point calls as part of its effort to drain user-generated point-to-point calls.  This is dangerous, and it's hard to guarantee that we wiol only drain user-generated point-to-point calls, and not some leftover MANA-generated point-to-point call.  So, fixing any bugs concerning forgotten MANA-generated point-to-point calls would become extremely difficult.

jungan:
> @jungan, The isssue is that collectiveToP2P is intended to convert all collective calls to P2P during normal _runtime_. Then, when it's time to checkpoint, MANA does:
> 
> 1. continue to execute until there are no collective calls in the lower half; and
> 2. then suspend all user threads, and use _only_ collective calls in an effort to drain any _user-generated_ pending point-to-point calls into buffers on the side of the receiver (the checkpoint thread should not be adding MANA-generated point-to-point calls that will be drained into buffers of user pending msg's), and then
> 3. do a checkpoint.
> 
> The problem with PR #80 is that after the user threads were suspended, it was converting MPI_Alltoall_internal to call the point-to-point versoin of MPI_Alltoall, which would then create MANA-generated point-to-point calls as part of its effort to drain user-generated point-to-point calls. This is dangerous, and it's hard to guarantee that we wiol only drain user-generated point-to-point calls, and not some leftover MANA-generated point-to-point call. So, fixing any bugs concerning forgotten MANA-generated point-to-point calls would become extremely difficult.

Oh, when I read the implementation, I thought `MPI_Alltoall_internal` will jump to the lower half, and calls real `MPI_Alltoall`, do you mean that in this case, it was called by `p2p_drain_send_recv.cpp` when checkpointing, and it won't jump to the lower half, but used the upper half `MPI_Alltoall` implementation which will be converted to P2P if defined `MPI_COLLECTIVE_P2P` ?

PR #85, opened 2022-03-23
Sequence number algorithm for checkpointing collective communications

xuyao0127:
This also includes cherry-picks from the collectiveToP2P branch.


xuyao0127:
@gc00 The algorithm is in the file seq_num.cpp. `commit_begin` and `commit_finish` are user thread functions that are called before and after real collective functions. The function `drainMpiCollectives` is the ckpt thread function that is called when DMTCP:PRESUSPEND event happens.

PR #87, opened 2022-03-25
[MANA-78] vdso was not restored properly on CentOS 7

jungan:
The root cause is:
1) We saved the temporary vvar&vdso addresses in global variables, the address of which could have been changed.
2) We lost vdso label after `mremap` it.

Found a PR in dmtcp https://github.com/dmtcp/dmtcp/pull/961/files made by @karya0  that handling the same thing. I cannot just cherry-pick it because there're some other differences. But I follow the idea there as much as possible.

Tested on my CentOS 7 and it's working fine now.

### DIFF
--- a/src/mtcp/mtcp_restart.c
+++ b/src/mtcp/mtcp_restart.c
@@ -499,25 +499,24 @@ mysetauxval(char **evp, unsigned long int type, unsigned long int val)
 /* We hardwire these, since we can't make libc calls. */
 #define PAGESIZE 4096
 #define ROUNDADDRUP(addr, size) ((addr + size - 1) & ~(size - 1))
-// We use global variables here, since both remap_vdso_and_vvar_regions and
-// unmap_memory_areas_and_restore_vdso use these variables. They're used to
-// track the temporary regions the vvar and vdso segments are remapped to, so
-// they are not unmapped later.
-uint64_t vvarStartTmp = 0;
-uint64_t vdsoStartTmp = 0;
 
 // This function searches for a free memory region to temporarily remap the vdso
 // and vvar regions to, so that mtcp's vdso and vvar do not overlap with the
 // checkpointed process' vdso and vvar.
 static void
-remap_vdso_and_vvar_regions() {
+remap_vdso_and_vvar_regions(RestoreInfo *rinfo) {
   Area area;
-  int rc = 0;
-  uint64_t vvarStart = 0;
-  uint64_t vdsoStart = 0;
-  uint64_t vvarSize = 0;
-  uint64_t vdsoSize = 0;
+  void *rc = 0;
+  uint64_t tmpVvarStart = 0;
+  uint64_t tmpVdsoStart = 0;
+  size_t vvarSize = 0;
+  size_t vdsoSize = 0;
   uint64_t prev_addr = 0x10000;
+
+  rinfo->currentVdsoStart = NULL;
+  rinfo->currentVdsoEnd = NULL;
+  rinfo->currentVvarStart = NULL;
+  rinfo->currentVvarEnd = NULL;
 
   int mapsfd = mtcp_sys_open2("/proc/self/maps", O_RDONLY);
   if (mapsfd < 0) {
@@ -527,14 +526,16 @@ remap_vdso_and_vvar_regions() {
 
   while (mtcp_readmapsline(mapsfd, &area)) {
     if (mtcp_strcmp(area.name, "[vvar]") == 0) {
-      vvarStart = area.addr;
+      rinfo->currentVvarStart = area.addr;
+      rinfo->currentVvarEnd = area.endAddr;
       vvarSize = area.size;
     } else if (mtcp_strcmp(area.name, "[vdso]") == 0) {
-      vdsoStart = area.addr;
+      rinfo->currentVdsoStart = area.addr;
+      rinfo->currentVdsoEnd = area.endAddr;
       vdsoSize = area.size;
     }
 
-    if (vvarStart > 0 && vdsoStart > 0) {
+    if (vvarSize > 0 && vdsoSize > 0) {
       break;
     }
   }
@@ -542,9 +543,13 @@ remap_vdso_and_vvar_regions() {
   mtcp_sys_lseek(mapsfd, 0, SEEK_SET);
 
   while (mtcp_readmapsline(mapsfd, &area)) {
-    if (prev_addr + vvarSize + vdsoSize <= area.addr) {
-      vvarStartTmp = prev_addr;
-      vdsoStartTmp = prev_addr + vvarSize;
+    if (prev_addr + vvarSize + vdsoSize <= area.__addr) {
+      if (vvarSize > 0) {
+        /* some kernel doesn't have vvar */
+        tmpVvarStart = prev_addr;
+      }
+      /* Assuming vdso always comes right after vvar */
+      tmpVdsoStart = prev_addr + vvarSize;
       break;
     } else {
       prev_addr = ROUNDADDRUP((uint64_t) area.endAddr, PAGESIZE);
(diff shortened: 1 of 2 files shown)


JainTwinkle:
@jungan There is a piece of code that skips the temporary vDSO and vvar regions while unmapping mtcp's own memory regions.
https://github.com/mpickpt/mana/blob/943ba896851adb2edf10fde74309f849ba247a2c/src/mtcp/mtcp_restart.c#L1444
```
    } else if (area.addr == vdsoStartTmpCopy) {
      DPRINTF("Skipping temporary vDSO section: %p-%p\n",
              area.addr, area.endAddr);
    } else if (area.addr == vvarStartTmpCopy) {
      DPRINTF("Skipping temporary vvar section: %p-%p\n",
              area.addr, area.endAddr);
    } 
```
Did you check why this piece of code doesn't work in your case?
It would be better to fix the code here than to add a new field in the rinfo structure.
What do you think @gc00?

jungan:
> @jungan There is a piece of code that skips the temporary vDSO and vvar regions while unmapping mtcp's own memory regions.
> 
> https://github.com/mpickpt/mana/blob/943ba896851adb2edf10fde74309f849ba247a2c/src/mtcp/mtcp_restart.c#L1444
> 
> ```
>     } else if (area.addr == vdsoStartTmpCopy) {
>       DPRINTF("Skipping temporary vDSO section: %p-%p\n",
>               area.addr, area.endAddr);
>     } else if (area.addr == vvarStartTmpCopy) {
>       DPRINTF("Skipping temporary vvar section: %p-%p\n",
>               area.addr, area.endAddr);
>     } 
> ```
> 
> Did you check why this piece of code doesn't work in your case? It would be better to fix the code here than to add a new field in the rinfo structure. What do you think @gc00?

It's because `vdsoStartTmpCopy` becomes 0 in this function, it because of 2 reasons.
1. Previously, we `mremap` vdso and save the temporary vdso address to a global variable `vdsoStartTmp`, the address of which will be changed after the following code:
```
  DPRINTF("We have copied mtcp_restart to higher address.  We will now\n"
          "    jump into a copy of restorememoryareas().\n");
  rinfo.vvarStartTmp = vvarStartTmp;
  rinfo.vdsoStartTmp = vdsoStartTmp;
  DPRINTF("%p %p %p %p\n", &vvarStartTmp, vvarStartTmp, &vdsoStartTmp, vdsoStartTmp);
  /* after calling the following assembly, the address of vvarStartTmp & vdsoStartTmp changed */

#if defined(__i386__) || defined(__x86_64__)
  asm volatile (CLEAN_FOR_64_BIT(mov %0, %%esp; )
```
2. In `unmap_memory_areas_and_restore_vdso` we will try the get the temporary vdso address again by parsing proc map, but the `vdso` label was dropped after we `mremap` it, so it doesn't work either.

Because of the combination of the above two, we lost `vdso` and didn't restore it correctly on my CentOS 7.9

JainTwinkle:
> In unmap_memory_areas_and_restore_vdso we will try the get the temporary vdso address again by parsing proc map, but the vdso label was dropped after we mremap it, so it doesn't work either.

I have two questions then:
1. Why are we searching for vdso and vvar if we already know that they are temporarily located at where vdsoStartTmp and vvarStartTmp point? 
2. Is the value of temporary vdso and vvar variables not persistent after we move to the hole?
    In other words, do vvarStartTmp and vdsoStartTmp become invalid (0) when mtcp unmap its own regions?

src/mtcp/mtcp_restart.c
```diff
@@ -1374,14 +1383,14 @@ unmap_memory_areas_and_restore_vdso(RestoreInfo *rinfo, LowerHalfInfo_t *lh_info
   /* Unmap everything except this image, vdso, vvar and vsyscall. */
   int mtcp_sys_errno;
   Area area;
-#if LINUX_VERSION_CODE > KERNEL_VERSION(3, 10, 0) 
+#if LINUX_VERSION_CODE > KERNEL_VERSION(3, 10, 0)
```
gc00, line 1386: @jungan , Shouldn't the `>`, above, be `>=`, and maybe even `>= KERNEL_VERSION(3, 11, 0)`?  At least, we wanted `>=`.

I thought that in our discussion, we saw that in Linux 3.10.x, the kernel stopped labeling vdso when we did mremap on it?  But the `#if` was testing only for `3.10.0` and not for `3.10.x`.  And I believe you were using a kernel with some version like 3.10.4, which was greater than 3.10.0, and yet the vdso label still was being lost after mremap.  And in that case, we decided to change this to:
```
#if LINUX_VERSION_CODE >= KERNEL_VERSION(3, 11, 0)
```

jungan: Hi @gc00 , I actually found that on my CentOS 7 env, the kernel version is 3.10.0, so it will try to use `vdsoStartTmp` instead of setting it to `NULL` and relies on parsing the proc map.

The problem is, `vdsoStartTmp` becomes 0 here. The place where it started to change is here:
```
  DPRINTF("We have copied mtcp_restart to higher address.  We will now\n"
          "    jump into a copy of restorememoryareas().\n");
  rinfo.vvarStartTmp = vvarStartTmp;
  rinfo.vdsoStartTmp = vdsoStartTmp;
  DPRINTF("%p %p %p %p\n", &vvarStartTmp, vvarStartTmp, &vdsoStartTmp, vdsoStartTmp);
  /* after calling the following assembly, the address of vvarStartTmp & vdsoStartTmp changed */

#if defined(__i386__) || defined(__x86_64__)
  asm volatile (CLEAN_FOR_64_BIT(mov %0, %%esp; )
```

And it's actually because the address of the global variable `vvarStartTmp ` changed.



In terms of shall we change to `#if LINUX_VERSION_CODE >= KERNEL_VERSION(3, 11, 0)`, I don't know since I didn't try if kernel 3.10.x has the label or not.

But like @JainTwinkle also asked, if we knew we have `mremap` the vdso and we saved the new address, is it necessary to still be purely relying on parsing the proc map? Even we think it's more reliable, but it doesn't make it worse if we just init the `vdsoStart` to the address we knew we have `mremap`, we can update it if we found the `vdso` in the proc map, but I think it's fine to just use the one we saved if we didn't find it in the proc map.

gc00:
And I have a comment for @JainTwinkle .  The bug being fixed here in MANA should be a concern for the latest DMTCP?  I don't see anything about `#if KERNEL_VERSION >= LINUX_KERNEL(3, 10, 0)` in the latest DMTCP.  Do we need to fix something similar in the latest DMTCP?  At the very least, if we do ckpt-restart-ckpt, then on the second ckpt, we can no longer find the 'vdso' label, and so we can't store the vdso label in the mtcpHeader during the second ckpt.  And so, the second restart can fail.

In general, we hit the bug here because @fyshhh fixed an earlier bug by first doing mremap to a new location, and then doing mremap to the final location.  Probably that fix was good, but then in Linux 3.10.0 and earlier, we lose the 'vdso' label.

Anyway, soon we will start using the latest DMTCP.  My largest concern is whether the latest DMTCP fixes this bug, or whether it fails on ckpt-restart-ckpt-restart.

I'm not sure which kernels fails to label 'vdso' after 'mremap'.  Eventually, the latest kernel does it right, but 3.10.0 does it wrong.

If you want to test various kernels, you can run this program with 'gdb' and `info proc mapping`:
```
#define _GNU_SOURCE
#include <sys/mman.h>
// Set these to correspond to the initial '[vdso]' on your Linux
#define VDSO_START 0x7ffff7ffa000 
#define VDSO_LENGTH 0x2000                                                                                                                                        int main() {
  mremap((void *)VDSO_START , VDSO_LENGTH , VDSO_LENGTH , MREMAP_MAYMOVE | MREMAP_FIXED, (void *)0x10000);
  return 0;
} 
```

gc00:
@jungan, This seems correct to me.  In less than a month, we should be replacing the 3-year old DMTCP from the original MANA with the latest DMTCP.  This will be accomplished when we merge in the `feature/dmtcp-master` branch being developed by @karya0 and @xuyao0127.  Naturally, we will first have to update that branch to include relevant PRs from the master branch, before we merge it in.

If this PR passes your tests, I suggest that we push it in now, and then examine the issue again when we merge in `feature/dmtcp-master`.

jungan:
> @jungan, This seems correct to me. In less than a month, we should be replacing the 3-year old DMTCP from the original MANA with the latest DMTCP. This will be accomplished when we merge in the `feature/dmtcp-master` branch being developed by @karya0 and @xuyao0127. Naturally, we will first have to update that branch to include relevant PRs from the master branch, before we merge it in.
> 
> If this PR passes your tests, I suggest that we push it in now, and then examine the issue again when we merge in `feature/dmtcp-master`.

Yes, it passed my test, but I don't understand why it can pass the scenario "ckpt-restart-ckpt-restart" you mentioned above, I confirmed after the restart, the process doesn't have the `vdso` in the proc map anymore, but somehow checkpoint still finds it and saves it in the image, confirmed by `readdmtcp.sh`, so it works fine to restart it again, which I don't know how it worked.

PR #89, opened 2022-03-29
Deterministic p2p mode - Fix several issues in logging requests

dahongli:
This change fixes several issues in logging requests for deterministic p2p mode.
[1] Fix open request log
[2] Logging requests for MPI_Test and MPI_Wait
[3] Initialize MPI before using its function in mana_p2p_update_log
[4] Remove infinite loop in mana_p2p_update_log.c

### DIFF
--- a/contrib/mpi-proxy-split/mpi-wrappers/mana_p2p_update_logs.c
+++ b/contrib/mpi-proxy-split/mpi-wrappers/mana_p2p_update_logs.c
@@ -15,6 +15,7 @@ void fill_in_log(struct p2p_log_msg *p2p_log);
 int main() {
   char buf[100];
   int rank;
+  MPI_Init(NULL, NULL);
   MPI_Comm_rank(MPI_COMM_WORLD, &rank);
   snprintf(buf, sizeof(buf)-1, P2P_LOG_MSG, rank);
   int fd_log = open(buf, O_RDWR);
@@ -37,6 +38,7 @@ int main() {
     lseek(fd_log, offset, SEEK_SET);
     writeall(fd_log, &p2p_log, sizeof(p2p_log));
   }
+  MPI_Finalize();
 
   return 0;
 }
--- a/contrib/mpi-proxy-split/mpi-wrappers/mana_p2p_update_logs.c
+++ b/contrib/mpi-proxy-split/mpi-wrappers/mana_p2p_update_logs.c
@@ -70,7 +70,7 @@ void fill_in_log(struct p2p_log_msg *p2p_log) {
     }
   }
   close(fd2);
-#ifdef 0
+#if 0
   // The following code has infinite loop. Comment out for now.
   fd2 = dup(fd_request);
   while (1) {
@@ -79,5 +79,5 @@ void fill_in_log(struct p2p_log_msg *p2p_log) {
       readall(fd_request, &p2p_request, sizeof(p2p_request));
     }
   }
-#endif // ifdef 0
+#endif // if 0
 }
--- a/contrib/mpi-proxy-split/mpi-wrappers/mana_p2p_update_logs.c
+++ b/contrib/mpi-proxy-split/mpi-wrappers/mana_p2p_update_logs.c
@@ -70,14 +70,4 @@ void fill_in_log(struct p2p_log_msg *p2p_log) {
     }
   }
   close(fd2);
-#if 0
-  // The following code has infinite loop. Comment out for now.
-  fd2 = dup(fd_request);
-  while (1) {
-    readall(fd2, &p2p_request, sizeof(p2p_request));
-    if (p2p_request.request == MPI_REQUEST_NULL) {
-      readall(fd_request, &p2p_request, sizeof(p2p_request));
-    }
-  }
-#endif // if 0
 }
--- a/contrib/mpi-proxy-split/mpi-wrappers/mana_p2p_update_logs.c
+++ b/contrib/mpi-proxy-split/mpi-wrappers/mana_p2p_update_logs.c
@@ -70,6 +70,8 @@ void fill_in_log(struct p2p_log_msg *p2p_log) {
     }
   }
   close(fd2);
+#ifdef 0
+  // The following code has infinite loop. Comment out for now.
   fd2 = dup(fd_request);
   while (1) {
     readall(fd2, &p2p_request, sizeof(p2p_request));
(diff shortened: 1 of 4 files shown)


contrib/mpi-proxy-split/mpi-wrappers/mana_p2p_update_logs.c
```diff
@@ -68,11 +70,14 @@ void fill_in_log(struct p2p_log_msg *p2p_log) {
     }
   }
   close(fd2);
+#ifdef 0
```
gc00, line 73: Pls change to `#if 0`  (Jenkins currently failing).

contrib/mpi-proxy-split/mpi-wrappers/mana_p2p_update_logs.c
```diff
@@ -68,11 +70,14 @@ void fill_in_log(struct p2p_log_msg *p2p_log) {
...
+  // The following code has infinite loop. Comment out for now.
   fd2 = dup(fd_request);
   while (1) {
     readall(fd2, &p2p_request, sizeof(p2p_request));
     if (p2p_request.request == MPI_REQUEST_NULL) {
       readall(fd_request, &p2p_request, sizeof(p2p_request));
```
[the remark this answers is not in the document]
dahongli: @gc00 The line 63-69 already find the same request matching the given p2p_log->request. Why do we need this loop from line 76-80. If we break when readall() return 0, then most likely we will keep reading until the end of the file. However, we did not use the read buffer p2p_request anymore after we break from this loop. Should we simply delete the loop of line 75 to line 80?

dahongli (commented):
Thanks Gene for your review!

contrib/mpi-proxy-split/mpi-wrappers/mpi_request_wrappers.cpp
```diff
@@ -64,6 +64,7 @@ USER_DEFINED_WRAPPER(int, Test, (MPI_Request*) request,
     return MPI_SUCCESS;
   }
   DMTCP_PLUGIN_DISABLE_CKPT();
+  LOG_PRE_Test(status);
```
gc00, line 67: Thanks for catching this missing macro.

Can we move this line one higher, to before `DMTCP_PLUGIN_DISABLE_CKPT();`?

It should be safe to allow checkpointing in the middle of this macro for logging.  DMTCP can safely save the open files associated with logging.  So, there's no reason to temporarily disable and delay checkpointing while we write to the log file.

contrib/mpi-proxy-split/mpi-wrappers/mpi_request_wrappers.cpp
```diff
@@ -210,6 +211,7 @@ USER_DEFINED_WRAPPER(int, Wait, (MPI_Request*) request, (MPI_Status*) status)
       fflush(stdout);
 #endif
     }
+    LOG_POST_Wait(request, statusPtr);
```
gc00, line 214: And similarly to above, we don't want to use log-replay here for MPI_Wait.  Please remove this.

contrib/mpi-proxy-split/mpi-wrappers/mana_p2p_update_logs.c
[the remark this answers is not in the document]
gc00: @dahongli, You're right.  When I was writing this code, I was assuming that we would log for a long time, but only replay for a short time (until we reach the issue related to a bug).  So, I never thought about what happens if we are replaying, and we reach the end of the log.

I'm not sure right now what is the right thing to do in this case.  We could just print a message saying that we have reached the end of the log, and then unset the environment variable `MANA_P2P_REPLAY`.

I'm not sure what is the correct behavior.  If you like the behavior that I suggest, please implement it.  Or feel free to propose a different behavior.

gc00 (commented):
@dahongli,  Thanks for this careful review.  Please seem my various inline comments.  The PR is almost there, and we can finish this off quickly, together.

contrib/mpi-proxy-split/mpi-wrappers/mana_p2p_update_logs.c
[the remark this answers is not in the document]
dahongli: This code updates log messages p2p_log_<rank>.txt bases on the receiving order of the requests from p2p_log_request_<rank>.txt. This update should happen before we replay the log messages in checkpointing restart. The general workflow I think something like
1. set `MANA_P2P_LOG`
2. launch application. This would create both message log file p2p_log_<rank>.txt and request log file p2p_log_request_<rank>.txt
3. Checkpointing
4. Let the application keep running until the application dies or stops
5. Now we run mana_p2p_update_log. At this moment, we have not restart from checkpointing image yet. 
6. Set replay environment variable `MANA_P2P_REPALY`
7. Restart checkpointing 

If we do the above workflow, this mana_p2p_update_log code only needs to read a request from request log and find the matching request in the message log and do the update. We do not need the loop from line 73 to 80. We could just delete them for now. 

The drawback is the log file records the sequence of all messages and receiving requests. It is not necessary. In the future, we could optimize it by dynamically enable or disable the logging of a running processes. We could this optimization later.

gc00 (commented):
@dahongli , Thanks for catching and addressing the issue when we reach the end of the log.

The only things remaining are two issues that I marked inline:
1. `LOG_PRE_Test(status);` : asking to move it one line higher
2. `LOG_PRE_Wait(status) : remove this line.  It will create a bug.
3. `LOG_POST_Wait(...)` : and remove this corresponding line.  It will create a bug.

After that, I can approve the PR, and it's ready to go.  (And I can push in PR #90, after your review it.)

contrib/mpi-proxy-split/mpi-wrappers/mpi_request_wrappers.cpp
dahongli: Currently only LOG_POST_Wait() takes the request status->{tag, source}. It is a good idea to remove this line. However, if we remove this line, then we have to change LOG_POST_Irecv to retrieve the {tag, source} from status and save it in the message log. If we go through this way, we do not need p2p_log_request_<rank>.txt created by save_request_info() function which is called by LOG_POST_Wait(). Further we could remove mana_p2p_update_log.c alltogether.

gc00 (approved):
Thanks.  It all looks good now.

PR #93, opened 2022-03-31
MANA follows --enable-debug

gc00:
MANA now uses the same default flags as for DMTCP.  So, `./configure --enable-debug` makes MANA use `-g -O0`.  Otherwise, MANA uses `-O2`.  (DMTCP does not set any flags for the FORTRAN compiler.)


dahongli (approved):
The performance of VASP5 PdO2 test run with option "--enable-debug" configuration is on par with that of default configuration. The detail result is below. 
==== VASP5 PdO2 test run ===
2 node, total 64 processes on Cori
Test Run,                   Total Run Time
Default.                          595 (s)
Enable-debug               589 (s)

The change looks good. Thanks Gene!

gc00:
@dahongli,
    Previously, the "default configuration" was essentially using `--enable-debug` (-g3 -O0).
    In this new PR, the default configuration (`./configure`) will now be using "-g -O2".

In your reported results, did you re-run the default configuration, based on using this branch, and re-compiling?
Thanks,
- Gene

PR #95, opened 2022-04-03
Update REPLAY_PRE_Irecv to replay the next message

dahongli:
This change updates REPLAY_PRE_Irecv to replay the next message and also fixes the utility functions it uses. It also enhance mana_update_p2p_log to show the messages saved in the log.


gc00:
@dahongli ,
  &nbsp;&nbsp;  There was a compile error. `p2p-deterministi.c` was declaring `get_next_message()` to return `int`, but `p2p-deterministic.txt` (which creates the '.h' file) was declaring it to return `void`.  I fixed that, and directly pushed it into your branch, so that we could see the Jenkins result.
&nbsp;&nbsp;   Jenkins is now passing.
&nbsp;&nbsp; Please do a `git pull --rebase YOUR_GITHUB_REPO bugfix` and test if the deterministic-p2p feature is now working inthat branch.
Thanks.

PR #96, opened 2022-04-07
[MANA-95] Add test case to reproduce the infinite loop issue

jungan:
There're 2 bugs in [MANA-95](https://memverge.atlassian.net/browse/MANA-95), one is fixed by PR #98 (PR #97 is the one to reproduce that issue).

This PR adds a test case to reproduce the issue that there might have another infinite loop in `completePendingIrecvs`

The root cause is when checkpoints, rank 1 is trying to receive a message from rank 0, but rank 0 hasn't sent it yet. So the send/receive bytes matched between rank 1 and rank 0, but there is one pending RECV request in rank 1's `g_async_calls`
Also at this moment, rank 2 has sent a message to rank 1, but rank 1 hasn't got a chance to receive it yet. that's why the send/receive bytes between rank 1 and rank 2 don't match, and we're hoping, the `recvFromAllComms` can recover this case.
But the rank 1 is stuck at `completePendingIrecvs`, and didn't get a chance to recover.

```
Rank 0:
int data = 0;
loop {
  sleep(10);
  MPI_Send(i++, 1);
  MPI_Barrier();
}

Rank 1:
loop {
  MPI_Recv(0);
  MPI_Recv(2);
  MPI_Barrier();
}

Rank 2:
int data = 0;
loop {
  MPI_Send(i++, 1);
  MPI_Barrier();
}
```

### DIFF
--- a/contrib/mpi-proxy-split/test/send_recv_loop.c
+++ b/contrib/mpi-proxy-split/test/send_recv_loop.c
@@ -1,0 +1,77 @@
+// Author: Wes Kendall
+// Copyright 2011 www.mpitutorial.com
+// This code is provided freely with the tutorials on mpitutorial.com. Feel
+// free to modify it for your own use. Any distribution of the code must
+// either provide a link to www.mpitutorial.com or keep this header in tact.
+//
+// MPI_Send, MPI_Recv example. Communicates the number -1 from process 0
+// to processe 1.
+//
+#include <mpi.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <assert.h>
+#include <unistd.h>
+#include <limits.h>
+#include <time.h>
+
+int main(int argc, char** argv) {
+  int iteration = 10;
+  if (argc > 1) {
+    iteration = atoi(argv[1]);
+  }
+
+  // Initialize the MPI environment
+  MPI_Init(NULL, NULL);
+  // Find out rank, size
+  int world_rank;
+  MPI_Comm_rank(MPI_COMM_WORLD, &world_rank);
+  int world_size;
+  MPI_Comm_size(MPI_COMM_WORLD, &world_size);
+
+  // We are assuming at least 2 processes for this task
+  if (world_size < 3) {
+    fprintf(stderr, "World size must be greater than or equal to 3 for %s\n", argv[0]);
+    MPI_Abort(MPI_COMM_WORLD, 1);
+  }
+
+  // rank 0 wait a while then send messages to rank 1
+  if (world_rank == 0) {
+    int number = 0;
+    for (int i = 0; i < iteration; i++) {
+      sleep(10);
+      MPI_Send(&number, 1, MPI_INT, 1, 0, MPI_COMM_WORLD);
+      number++;
+      MPI_Barrier(MPI_COMM_WORLD);
+      printf("Rank 0 have successfully sent %d messages to rank 1\n", i + 1);
+      fflush(stdout);
+    }
+  } else if (world_rank == 1) {
+    // rank 1 receive message from 0 first and then 2
+    int number = 0;
+    for (int i = 0; i < iteration; i++) {
+      int recv_number = -1;
+      MPI_Recv(&recv_number, 1, MPI_INT, 0, 0, MPI_COMM_WORLD, MPI_STATUS_IGNORE);
+      assert(number == recv_number);
+      MPI_Recv(&recv_number, 1, MPI_INT, 2, 0, MPI_COMM_WORLD, MPI_STATUS_IGNORE);
+      assert(number == recv_number);
+      number++;
+      MPI_Barrier(MPI_COMM_WORLD);
+      printf("Rank 1 have successfully received %d messages\n", (i + 1) * 2);
+      fflush(stdout);
+    }
+  } else {
+    // rank 2 send messages to rank 1 right away
+    int number = 0;
+    for (int i = 0; i < iteration; i++) {
+      MPI_Send(&number, 1, MPI_INT, 1, 0, MPI_COMM_WORLD);
+      number++;
+      MPI_Barrier(MPI_COMM_WORLD);
+      printf("Rank 2 have successfully sent %d messages to rank 1\n", i + 1);
+      fflush(stdout);
+    }
+
+  }
+  MPI_Finalize();
+  return 0;
+}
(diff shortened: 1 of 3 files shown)


gc00 (commented):
@jungan , Probably it wasn't clear from the e-mail of @xuyao0127.  But his file `p2p_bug_fix.txt` already had the desired regression test.  @JainTwinkle, @xuyao0127 and I chose this regresssion test:
```

Regression Test:

Make this standard Jenkins test. Choose number of iterations for convenience.

Rank 1:
    int i = 0;
    loop {
        // Same tag and comm for all sends
        MPI_Send(i++)
        MPI_Send(i++)
        MPI_Send(i++)
        sleep(5)
        MPI_Barrier()
    }

Rank 2:
    loop {
        sleep(5)
        // Also verify if messages are in order
        MPI_Recv()
        MPI_Recv()
        MPI_Recv()
        MPI_Barrier()
    }
```

The motivation is that after each barrier, the sender sends three times and then sleeps.  But the receiver sleeps first, and then receives three times.  As a result, any ckpt is most likely to occur during the sleep.  So, a ckpt occurs _after_ the three sends and _before_ the three receives.

In your regression test, you are simply doing many sends and receives, and hoping that statistically, after many checkpoints, we will eventually see the bug.  But the regression test of @xuyao0127 will exhibit the bug even the first time that there is a checkpoint.  So, his regression test is preferred.  It's also a good thing that the regression test of @xuyao0127 works with just two ranks.

Of course, if you want it to work when there are many ranks, then you would have to add something like:
```
// Before entering the loop
color = world_rank < 2;  // Assuming you already computed the world_rank
MPI_Comm comm_with_two_ranks;
MPI_Comm_split(MPI_COMM_WORLD, color, world_rank, &comm_with_two_ranks);
// After entering the loop
if (rank == 0) { // if sender
  ... // send and sleep
  MPI_Barrier(comm_with_two_ranks);
else if (rank == 1) { // if receiver
  MPI_Barrier(comm_with_two_ranks);
  ...  // sleep and recv
```

Or if you prefer, you could first create an explicit group, and then the communicator.  But the paradigm above is shorter.

jungan:
Sorry, I misunderstood it, now I think I fully understand it now, although my test case is able to reproduce the issue, it's certainly better that we can reproduce it at every checkpoint.
The reason why I make it support testing more ranks is that I'm hoping to reproduce the "first" infinite loop, as I only saw the second one when testing it with two ranks.
After a second thought, I think it would be better to only test one scenario in one test case, I will update the test according to your suggestion, and try to write another test case to reproduce the other infinite loop. (Or after I got @xuyao0127 's fix and try that with the wave_mpi first)

> @jungan , Probably it wasn't clear from the e-mail of @xuyao0127. But his file `p2p_bug_fix.txt` already had the desired regression test. @JainTwinkle, @xuyao0127 and I chose this regresssion test:
> 
> ```
> 
> Regression Test:
> 
> Make this standard Jenkins test. Choose number of iterations for convenience.
> 
> Rank 1:
>     int i = 0;
>     loop {
>         // Same tag and comm for all sends
>         MPI_Send(i++)
>         MPI_Send(i++)
>         MPI_Send(i++)
>         sleep(5)
>         MPI_Barrier()
>     }
> 
> Rank 2:
>     loop {
>         sleep(5)
>         // Also verify if messages are in order
>         MPI_Recv()
>         MPI_Recv()
>         MPI_Recv()
>         MPI_Barrier()
>     }
> ```
> 
> The motivation is that after each barrier, the sender sends three times and then sleeps. But the receiver sleeps first, and then receives three times. As a result, any ckpt is most likely to occur during the sleep. So, a ckpt occurs _after_ the three sends and _before_ the three receives.
> 
> In your regression test, you are simply doing many sends and receives, and hoping that statistically, after many checkpoints, we will eventually see the bug. But the regression test of @xuyao0127 will exhibit the bug even the first time that there is a checkpoint. So, his regression test is preferred. It's also a good thing that the regression test of @xuyao0127 works with just two ranks.
> 
> Of course, if you want it to work when there are many ranks, then you would have to add something like:
> 
> ```
> // Before entering the loop
> color = world_rank < 2;  // Assuming you already computed the world_rank
> MPI_Comm comm_with_two_ranks;
> MPI_Comm_split(MPI_COMM_WORLD, color, world_rank, &comm_with_two_ranks);
> // After entering the loop
> if (rank == 0) { // if sender
>   ... // send and sleep
>   MPI_Barrier(comm_with_two_ranks);
> else if (rank == 1) { // if receiver
>   MPI_Barrier(comm_with_two_ranks);
>   ...  // sleep and recv
> ```
> 
> Or if you prefer, you could first create an explicit group, and then the communicator. But the paradigm above is shorter.

Sorry, I misunderstood it, now I think I fully understand it now, although my test case is able to reproduce the issue, it's certainly better that we can reproduce it at every checkpoint.
The reason why I make it support testing more ranks is that I'm hoping to reproduce the "first" infinite loop, as I only saw the second one when testing it with two ranks.
After a second thought, I think it would be better to only test one scenario in one test case, I will update the test according to your suggestion, and try to write another test case to reproduce the other infinite loop. (Or after I got @xuyao0127 's fix and try that with the wave_mpi first)

gc00:
Yes, that works well:  keeping this test and also creating a new one targeted toward this specific bug.  Let's do that.

jungan:
> There are several files included in the PR are not related to the test case. Maybe you can put them in a separate PR.

OK, updated PR only includes the teat case, will have another PR for those small improvements.

contrib/mpi-proxy-split/test/send_recv_loop.c
```diff
@@ -0,0 +1,77 @@
...
+
+  // rank 0 wait a while then send messages to rank 1
+  if (world_rank == 0) {
+    int number = 0;
+    for (int i = 0; i < iteration; i++) {
+      sleep(10);
+      MPI_Send(&number, 1, MPI_INT, 1, 0, MPI_COMM_WORLD);
+      number++;
+      MPI_Barrier(MPI_COMM_WORLD);
+      printf("Have successfully sent %d messages\n", i + 1);
```
dahongli, line 46: Could you print current rank number  so we can know which rank is sending this message? Same for several other  printf.

dahongli (approved):
Thanks Jun for catching this issue! The  change looks good.

contrib/mpi-proxy-split/test/send_recv_loop.c
jungan: Added

PR #97, opened 2022-04-11
[MANA-95] Add test case to reproduce p2p infinite loop at ckpt

xuyao0127:
This test case is what I described in the email. The sender will send messages with the same length and same destination multiple times before the checkpoint. The receiver will receive these messages after the checkpoint. This is related to PR #96.


jungan (commented):
Any essential difference with PR #96 ?

xuyao0127:
@jungan Sorry, I didn't notice you updated the test case in PR #96. After reading the conversation, I thought you want to keep the original test case and create a new one. So I create this test case to test my fix. This is the same as PR #96. We can keep either one.

jungan (approved):

PR #98, opened 2022-04-11
[MANA-95] P2p bugfix

xuyao0127:
Currently, the MPI_Irecv wrapper, which will be called when draining p2p messages, will check the internal buffer used for draining during checkpoint time. As a result, if two messages need to be drained during checkpoint time and they have the same destination, tag, and size, MANA will try to "consume" the first message in the internal buffer and put it back into the internal buffer as the second message. The real second message will never be properly received. MPI_Iprobe will always remind MANA there's a pending message in the network, and the draining process for the second message will be repeated infinitely. Therefore, the p2p draining becomes an infinite loop.

The bug can be reproduced with the test case added in PR #97.

This fix adds a global variable `mana_state`, which shows the current state of MANA. Previously, a macro `MPI_LOGGING` has similar functionality, but can only distinguish the replay state during a restart. `mana_state` enables MANA to distinguish more states during its lifetime. Now, the `MPI_Irecv` wrapper will only check the internal draining buffer if `mana_state == RUNNING`.

### DIFF
--- a/contrib/mpi-proxy-split/mpi_plugin.cpp
+++ b/contrib/mpi-proxy-split/mpi_plugin.cpp
@@ -21,6 +21,7 @@
 
 #include <signal.h>
 
+#include "mpi_plugin.h"
 #include "lower_half_api.h"
 #include "split_process.h"
 #include "p2p_log_replay.h"
@@ -42,6 +43,7 @@ using namespace dmtcp;
 
 int g_numMmaps = 0;
 MmapInfo_t *g_list = NULL;
+mana_state_t mana_state = UNKNOWN_STATE;
 
 // #define DEBUG
 
@@ -125,6 +127,7 @@ mpi_plugin_event_hook(DmtcpEvent_t event, DmtcpEventData_t *data)
     {
       JTRACE("*** DMTCP_EVENT_INIT");
       initialize_segv_handler();
+      mana_state = RUNNING;
       JASSERT(!splitProcess()).Text("Failed to create, initialize lower haf");
       break;
     }
@@ -159,7 +162,36 @@ updateLhEnviron()
   fnc(__environ);
 }
 
+/* ===============================================================
+ * Helper functions that sets mana_state. These functions are used
+ * in mpiPluginBarriers according to different events.
+ *
+ * FIXME: In feature/dmtcp-master, all events are handled in the
+ * mpi_plugin_event_hook() function. In that case, we can use
+ * functions with argument. Therefore, we can merge all of these
+ * helper functions into one function: set_mana_state(state).
+ */
+static void setManaStateRunning() {
+  mana_state = RUNNING;
+}
+
+static void setManaStateCkptCollective() {
+  mana_state = CKPT_COLLECTIVE;
+}
+
+static void setManaStateCkptP2p() {
+  mana_state = CKPT_P2P;
+}
+
+static void setManaStateRestartReplay() {
+  mana_state = RESTART_REPLAY;
+}
+/* =============================================================== */
+
+
 static DmtcpBarrier mpiPluginBarriers[] = {
+  { DMTCP_GLOBAL_BARRIER_PRE_SUSPEND, setManaStateCkptCollective,
+    "set-mana-state-ckpt-collective"},
   { DMTCP_GLOBAL_BARRIER_PRE_SUSPEND, NULL,
     "Drain-MPI-Collectives", drainMpiCollectives},
   { DMTCP_PRIVATE_BARRIER_PRE_CKPT, logIbarrierIfInTrivBarrier,
@@ -170,6 +202,8 @@ static DmtcpBarrier mpiPluginBarriers[] = {
     "GetLocalRankInfo"},
   { DMTCP_GLOBAL_BARRIER_PRE_CKPT, updateCkptDirByRank,
     "update-ckpt-dir-by-rank" },
+  { DMTCP_GLOBAL_BARRIER_PRE_CKPT, setManaStateCkptP2p,
+    "set-mana-state-ckpt-p2p" },
   { DMTCP_GLOBAL_BARRIER_PRE_CKPT, registerLocalSendsAndRecvs,
     "Register-local-sends-and-receives" },
   { DMTCP_GLOBAL_BARRIER_PRE_CKPT, drainSendRecv,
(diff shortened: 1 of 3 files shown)


xuyao0127:
This fix has been tested with the test case added in PR #97. But Jenkins shows some building or unit test error. I don't have access to Jenkins, @gc00 @jungan can you help me to find what is the error. It will be better if I can also have the access to Jenkins.

contrib/mpi-proxy-split/mpi-wrappers/mpi_wrappers.cpp
```diff
@@ -47,6 +48,7 @@ USER_DEFINED_WRAPPER(int, Init, (int *) argc, (char ***) argv) {
 }
 USER_DEFINED_WRAPPER(int, Init_thread, (int *) argc, (char ***) argv,
                      (int) required, (int *) provided) {
+  mana_state = RUNNING;
```
jungan, line 51: So before the user application calls `MPI_Init` or `MPI_Init_thread`, the `mana_state` is `UNKNOWN_STATE`?
But we set `mana_state` to `RUNNING` before it actually calls the real `MPI_Init`?

In my mind, we either set `mana_state` to `RUNNING` as soon as the process started, or we should do it after the real `MPI_Init`.

It may don't matter for now, but I suggest we have a more concrete design.

Two more questions:
1. Are those two places the only places that set `mana_state` in the user thread? It may not have race logically, just want to confirm.
2. I guess it should not matter if the user checkpoint the application before it calls `MPI_Init`?

jungan (commented):
Not sure why CI build is failing, here is the error in case you cannot see it:
```
18:54:12 cc -o Alloc_mem.mana.exe Alloc_mem.o  -L$PWD/../mpi-wrappers -lmpistub
18:54:12 rm Ibarrier_test.o Sendrecv_test.o Bcast_test.o Alltoallv_test.o Reduce_test.o Abort_test.o Comm_split_test.o Allreduce_test.o send_recv.o Waitall_test.o Alloc_mem.o Allgather_test.o mpi_hello_world.o Barrier_test.o Comm_dup_test.o send_recv_many.o Ibcast_test.o Group_size_rank.o Irecv_test.o Isend_test.o ping_pong.o Alltoall_test.o keyval_test.o hello_mpi_init_thread.o two-phase-commit-1.o Type_commit_contiguous.o two-phase-commit-2.o
18:54:12 make[4]: Leaving directory '/home/jenkins/workspace/mana_Pull_Request_Builder/contrib/mpi-proxy-split/test'
18:54:12 make[3]: Leaving directory '/home/jenkins/workspace/mana_Pull_Request_Builder/contrib/mpi-proxy-split/test'
18:54:12 make[2]: Leaving directory '/home/jenkins/workspace/mana_Pull_Request_Builder/contrib/mpi-proxy-split'
18:54:12 make[1]: Leaving directory '/home/jenkins/workspace/mana_Pull_Request_Builder'
18:54:12 + '[' '!' -f bin/lh_proxy -o '!' -f bin/mana_launch ']'
18:54:12 + cd /home/jenkins/workspace/mana_Pull_Request_Builder/ci/../contrib/mpi-proxy-split/unit-test
18:54:12 + make
18:54:12 mpic++ -g3 -O0 -fPIC -I../../../include -I.. -I../../../jalib -I../lower-half -I../mpi-wrappers -I../../../src -std=c++11 -D'NEXT_FUNC(fnc)=MPI_##fnc' -g3 -O0 -c -o record-replay-comm-test.o record-replay-comm-test.cpp
18:54:12 In file included from record-replay-comm-test.cpp:6:
18:54:12 ../virtual-ids.h:42: warning: "NEXT_FUNC" redefined
18:54:12  # define NEXT_FUNC(func)                                                       \
18:54:12  
18:54:12 <command-line>: note: this is the location of the previous definition
18:54:14 mpic++ -fPIC -g3 -O0 -o record-replay-comm-test.exe record-replay-comm-test.o ../record-replay.o ../../../src/libjalib.a ../../../src/libdmtcpinternal.a ../../../src/libnohijack.a -lgtest -lpthread
18:54:14 ../record-replay.o: In function `restoreMpiLogState()':
18:54:14 /home/jenkins/workspace/mana_Pull_Request_Builder/contrib/mpi-proxy-split/record-replay.cpp:72: undefined reference to `mana_state'
18:54:14 collect2: error: ld returned 1 exit status
18:54:14 make: *** [Makefile:52: record-replay-comm-test.exe] Error 1
18:54:14 rm record-replay-comm-test.o
18:54:14 + exit 1
18:54:14 Build step 'Execute shell' marked build as failure
18:54:14 Setting status of fa5da466d2954e6699fc31deafa57281c7bd0cd0 to FAILURE with url https://104.184.156.164:8888/job/mana_Pull_Request_Builder/260/ and message: 'Build finished. '
18:54:14 Using context: CI Build and Unit Test 
18:54:14 Finished: FAILURE
```

And here is my understanding CMIIWW:

Currently, we only checked `mana_stat` at one place, others are placeholder for now.

contrib/mpi-proxy-split/mpi-wrappers/mpi_wrappers.cpp
xuyao0127: I see. It should be my mistake initializing the variable. I will fix it today.

xuyao0127: I found the problem is the variable `mana_state` is defined in `mpi_plugin.cpp`, and declared in `mpi_plugin.h` as an extern variable. Usually `mpi_plugin.o` is compiled with `record-replay.o`. CI fails because the unit test `record-replay-comm-test.exe` only use `record-replay.o`, therefore the variable is only declared, but not defined. 

This fix update `mana_state` in different functions for each states (RUNNING, CKPT_COLLECTIVE, CKPT_P2P, etc.). Maybe a better way is using DMTCP event. This will be less intrusive. I will update this fix today.

xuyao0127:
@gc00 @jungan This new fix passed CI.

contrib/mpi-proxy-split/mpi_plugin.cpp
```diff
@@ -159,7 +162,26 @@ updateLhEnviron()
...
+  mana_state = CKPT_COLLECTIVE;
+}
+
+void setManaStateCkptP2p() {
+  mana_state = CKPT_P2P;
+}
+
+void setManaStateRestartReplay() {
+  mana_state = RESTART_REPLAY;
+}
```
jungan, line 179: NIT: consider making those functions static.

jungan (approved):

contrib/mpi-proxy-split/mpi_plugin.cpp
xuyao0127: Have made them static.

PR #99, opened 2022-04-12
[MANA-95] Fix: completePendingIrecvs has infinite loop

xuyao0127:
The for loop inside the function completePendingIrecvs doesn't update
  the iterator if MPI_Test fails. If the sender only sends the message
  after the checkpoint, this will be an infinite loop.

### DIFF
--- a/contrib/mpi-proxy-split/p2p_drain_send_recv.cpp
+++ b/contrib/mpi-proxy-split/p2p_drain_send_recv.cpp
@@ -136,6 +136,24 @@ completePendingIrecvs()
         g_recvBytesByRank[worldRank] += call->count * size;
         bytesReceived += call->count * size;
         it = g_async_calls.erase(it);
+      } else {
+        /*  We update the iterator even if the MPI_Test fails.
+         * Otherwise, the message we are waiting for will be sent
+         * after the checkpoint. This can result in an infinite loop.
+         *
+         * NOTE: This function will be called only if the global arrays
+         * do not match. This can happen if a second sender has sent
+         * a message to us, and we will receive the message only
+         * after the checkpoint. The following diagram is an example:
+         *
+         * RANK 0           RANK 1              RANK 2        TIME
+         *                                    Send to Rank 1   |
+         *                Recv from Rank 0                     |
+         * =====CKPT=======CKPT======CKPT======CKPT========    |
+         *                Recv from Rank 2                     |
+         * Send to Rank 1                                      V
+         */
+        it++;
       }
     } else {
       it++;


contrib/mpi-proxy-split/p2p_drain_send_recv.cpp
```diff
@@ -136,6 +136,14 @@ completePendingIrecvs()
         g_recvBytesByRank[worldRank] += call->count * size;
         bytesReceived += call->count * size;
         it = g_async_calls.erase(it);
+      } else {
+        // If the MPI_Test failed, we still update the iterator.
```
JainTwinkle, line 140: You might want to update the comment for clarity. Here's a suggested comment:
 ```
 /* We update the iterator even if the MPI_Test fails.
  * Otherwise, the message we are waiting for will be sent
  * after the checkpoint. This can result in an infinite loop.
  *
  * NOTE: This function will be called only if the global arrays
  * do not match. This can happen if a second sender has sent
  * a message to us, and we will receive the message only
  * after the checkpoint.
```

File: contrib/mpi-proxy-split/p2p_drain_send_recv.cpp [hunk 2]
```diff
@@ -136,6 +136,14 @@ completePendingIrecvs()
...
+        // Otherwise, if the message we are waiting for will be sent after
+        // the checkpoint, this can be an infinite loop.
+        // NOTE: This function will be called only if the global arrays do not
+        // match. This can happen if a second sender has sent to us, and we will
+        // receive only after the checkpoint.
+        it++;
```
jungan, line 146: I actually thought about this solution.
Is there a chance that the peer has sent the message, but the first time `MPI_Test_internal` failed and we want to retry?
I check the implementation of `MPI_Recv`, which calls `MPI_Wait`, I guess the reason it uses a loop over there is that we don't know when the peer sent the message?

jungan: On second thought, it doesn't matter, if that really happens, it will get back to here because of `while (!allDrained())`

jungan (approved):

contrib/mpi-proxy-split/p2p_drain_send_recv.cpp
```diff
@@ -136,6 +136,14 @@ completePendingIrecvs()
         g_recvBytesByRank[worldRank] += call->count * size;
         bytesReceived += call->count * size;
         it = g_async_calls.erase(it);
+      } else {
+        // If the MPI_Test failed, we still update the iterator.
+        // Otherwise, if the message we are waiting for will be sent after
+        // the checkpoint, this can be an infinite loop.
+        // NOTE: This function will be called only if the global arrays do not
+        // match. This can happen if a second sender has sent to us, and we will
+        // receive only after the checkpoint.
```
gc00, line 145: After @xuyao0127 and I had written this comment I thought that a small diagram might be helpful.  Perhaps something like:
```
Rank 0           Rank 1                        Rank 2            TIME
                                                      Send to Rank 1      |
                       Recv from Rank 0                                 |
=====CKPT==CKPT=======CKPT======                                         |
Send to Rank 1   Recv from Rank 2                                      V
```
(Needs better formatting)

gc00: @xuyao0127, Can we insert the diagram that I'm proposing into the comment, before we push this in?  As the saying goes, "A picture is worth a thousand words."

I'll try to clean up my example diagram.  The problem is figuring out how to display it correctly in a mono-spaced font.

xuyao0127: @gc00 I found this discussion about creating ASCII diagrams: https://unix.stackexchange.com/questions/126630/creating-diagrams-in-ascii. Hope it's helpful.

gc00: @xuyao0127 , Thanks for the pointer to ASCII diagrams.  It will be useful in the future.
For now, I've cleaned up my ASCII diagram.  You can just copy-paste it into the C++ file,
and prepend `// ` to make it a comment.

PR #100, opened 2022-04-18
Several small improvements

jungan:
- Make sure `make distclean` can clean up everything in mpi-proxy-split
- Check error returns
- Make some functions static when applicable
- Free pointers in `g_async_calls`
- Remove unused function & variable definitions


contrib/mpi-proxy-split/split_process.cpp
```diff
@@ -247,12 +247,10 @@ read_lh_proxy_bits(pid_t childpid)
...
   //        with ESRCH (no such proc.); it may need CAP_SYS_PTRACE privilege??
-  for (int i = 0; i < IOV_SZ; i++) {
-    JTRACE("Reading segment from lh_proxy")
-          (remote_iov[i].iov_base)(remote_iov[i].iov_len);
-    ret = process_vm_readv(childpid, remote_iov + i, 1, remote_iov + i, 1, 0);
-    JASSERT(ret != -1)(JASSERT_ERRNO).Text("Error reading data from lh_proxy");
-  }
+  JTRACE("Reading segment from lh_proxy")
+         (remote_iov[0].iov_base)(remote_iov[0].iov_len);
+  ret = process_vm_readv(childpid, remote_iov, IOV_SZ, remote_iov, IOV_SZ, 0);
```
jungan, line 252: @rohgarg @gc00 @xuyao0127 Not sure if I understand it correctly, I tested it and works fine.
Please comment if there is anything I missed.

contrib/mpi-proxy-split/p2p_drain_send_recv.cpp
```diff
@@ -97,8 +97,11 @@ recvMsgIntoInternalBuffer(MPI_Status status, MPI_Comm comm)
   MPI_Type_size(MPI_BYTE, &size);
   JASSERT(size == 1);
   void *buf = JALLOC_HELPER_MALLOC(count);
-  MPI_Recv(buf, count, MPI_BYTE, status.MPI_SOURCE, status.MPI_TAG,
+  int rc = MPI_Recv(buf, count, MPI_BYTE, status.MPI_SOURCE, status.MPI_TAG,
            comm, MPI_STATUS_IGNORE);
+  if (rc != MPI_SUCCESS) {
+    fprintf(stderr, "MPI_Recv failed\n");
```
rohgarg, line 103: Do we want to use JTRACE here?

contrib/mpi-proxy-split/p2p_drain_send_recv.cpp
```diff
@@ -135,6 +138,7 @@ completePendingIrecvs()
                                               call->comm);
         g_recvBytesByRank[worldRank] += call->count * size;
         bytesReceived += call->count * size;
+        JALLOC_HELPER_FREE(call);
```
rohgarg, line 141: I'll have to check, but IIRC, the call object couldn't be freed here because it was being referenced in some other place later. So, I had decided to live with the memory leak. I'll have to go through the code again to make sure.

contrib/mpi-proxy-split/p2p_drain_send_recv.cpp
```diff
@@ -190,17 +194,16 @@ recvFromAllComms()
...
 {
   dmtcp::map<MPI_Request, mpi_async_call_t*>::iterator it;
   for (it = g_async_calls.begin(); it != g_async_calls.end();) {
     MPI_Request request = it->first;
     mpi_async_call_t *call = it->second;
-    int flag = 0;
     if (call->type == ISEND_REQUEST) {
       UPDATE_REQUEST_MAP(request, MPI_REQUEST_NULL);
-      // FIXME: We should free `call' to avoid memory leak
+      JALLOC_HELPER_FREE(call);
```
rohgarg, line 206: Same as above

contrib/mpi-proxy-split/split_process.cpp
rohgarg: What's the intent here?

contrib/mpi-proxy-split/p2p_drain_send_recv.cpp
jungan: In this case, shall we use a `shared_ptr`

contrib/mpi-proxy-split/p2p_drain_send_recv.cpp
jungan: Sure, will update the code.

contrib/mpi-proxy-split/split_process.cpp
jungan: I think it would be better to take the advantage of `process_vm_readv`?
I know it doesn't have too much difference given `IOV_SZ` is 2 here, I think using the standard way probably performs better than we do the loop here when `IOV_SZ` is large.

contrib/mpi-proxy-split/split_process.cpp
```diff
@@ -333,7 +331,7 @@ startProxy()
       // Read from stdout of lh_proxy full lh_info struct, including orig memRange.
       close(pipefd_out[1]); // close write end of pipe
       // FIXME: should be a readall. Check for return error code.
-      if (read(pipefd_out[0], &lh_info, sizeof lh_info) < sizeof lh_info) {
+      if (read(pipefd_out[0], &lh_info, sizeof lh_info) < (ssize_t) sizeof lh_info) {
```
gc00, line 334: Please do not commit the change o this line.

The issue here is that `read` returns an `int` and `sizeof` returns a `size_t`. (unsigned).  Some compiler complain or warn about comparing an `int` and `unsigned int`.  The type `ssize_t` stands for "signed size", which is almost the same thing as an `int`.  So, this cast is needed to make some compilers stop issuing warnings.

gc00 (commented):
It seems like @rohgarg and @jungan have a continuing conversation here.  So, I'll add my review solely as a "comment".  @jungan said that he intends to provide an updated version of this PR later.

gc00:
@jungan and @rohgarg ,
    Where are we on this PR?  Shall we split off a portion of this that we all agree is useful, and then leave the rest for further consideration?
Best,
- Gene

jungan:
> @jungan and @rohgarg , Where are we on this PR? Shall we split off a portion of this that we all agree is useful, and then leave the rest for further consideration? Best,
> 
> * Gene

Firstly I think I need to rebase the code on top of `feature/dmtcp-master`, then I will dive deeper into the issue @rohgarg mentioned and see what's the best solution.

PR #101, opened 2022-04-19
Fix bug in libproxy.c where PID may be longer than expected

fyshhh:
This pull request fixes a bug where a `lh_proxy` process may have a PID longer than 6 digits, which would cause the end of the `pids` string to be cut off. This is an issue on CentOS Stream, where the max PID is 4194304.

### DIFF
--- a/contrib/mpi-proxy-split/lower-half/libproxy.c
+++ b/contrib/mpi-proxy-split/lower-half/libproxy.c
@@ -110,7 +110,9 @@ getTextSegmentRange(pid_t proc,                 // IN
   if (proc == -1) {
     f = fopen("/proc/self/stat", "r");
   } else {
-    char pids[] = "/proc/XXXXXX/stat";
+    // On 64-bit systems, pid_max can be set to any value up to 2^22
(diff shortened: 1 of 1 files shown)


contrib/mpi-proxy-split/lower-half/libproxy.c
```diff
@@ -110,7 +110,7 @@ getTextSegmentRange(pid_t proc,                 // IN
   if (proc == -1) {
     f = fopen("/proc/self/stat", "r");
   } else {
-    char pids[] = "/proc/XXXXXX/stat";
+    char pids[] = "/proc/XXXXXXX/stat";
```
jungan, line 113: Please add a comment:

On 64-bit systems, pid_max can be set to any value up to 2^22 (PID_MAX_LIMIT, approximately 4 million)

NIT: I would just use `char pids[PATH_MAX]`

jungan (approved):

contrib/mpi-proxy-split/lower-half/libproxy.c
fyshhh: both ideas sound good..

PR #102, opened 2022-04-21
Deterministic p2p - log and replay MPI_[Irecv,Iprobe,Wait,Test]

dahongli:
This change adds several enhancements and bug fixes in deterministic p2p log and replay. 

The main changes are,
- Log in MPI_Irecv, MPI_Iprobe, MPI_Wait and MPI_Test
- Don't truncate file to the saved file size saved in checkpoint image. The message log file continues saving messages post checkpoint while the program runs. So its size is larger.
- Change get_next_message to support when both MANA_P2P_LOG and MANA_P2P_REPLAY are enabled at the same time.
- Enhance mana_p2p_update_logs to handle the request number out of order.

### DIFF
--- a/contrib/mpi-proxy-split/mpi-wrappers/mana_p2p_update_logs.c
+++ b/contrib/mpi-proxy-split/mpi-wrappers/mana_p2p_update_logs.c
@@ -5,16 +5,25 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <mpi.h>
+#include <string.h>
 // To support MANA_P2P_LOG and MANA_P2P_REPLAY:
 #define USE_READALL
 #define USE_WRITEALL
 #include "p2p-deterministic.h"
 
 void fill_in_log(struct p2p_log_msg *p2p_log);
+int show_log(char *name);
 
-int main() {
+int main(int argc, char *argv[]) {
   char buf[100];
   int rank;
+  int rc;
+
+  if (argc == 2) {
+    rc = show_log(argv[1]);
+    return rc;
+  }
+
   MPI_Init(NULL, NULL);
   MPI_Comm_rank(MPI_COMM_WORLD, &rank);
   snprintf(buf, sizeof(buf)-1, P2P_LOG_MSG, rank);
@@ -28,7 +37,7 @@ int main() {
   while (1) {
     struct p2p_log_msg p2p_log;
     off_t offset = lseek(fd_log, 0, SEEK_CUR);
-    int rc = readall(fd_log, &p2p_log, sizeof(p2p_log));
+    rc = readall(fd_log, &p2p_log, sizeof(p2p_log));
     if (rc == 0) {
       break;
     }
@@ -71,3 +80,25 @@ void fill_in_log(struct p2p_log_msg *p2p_log) {
   }
   close(fd2);
 }
+
+int show_log(char *name)
+{
+  int fd_log = open(name, O_RDWR);
+  if (fd_log == -1) {
+    fprintf(stderr, "show_log open: fle %s, error: %s\n", name, strerror(errno)); 
+    return 1; 
+  }
+
+  while (1) {
+    struct p2p_log_msg p2p_log;
+    int rc = readall(fd_log, &p2p_log, sizeof(p2p_log));
+    if (rc == 0) {
+      break;
+    }
+
+    printf("MSG(request,source,tag,count,comm): %d,%d,%d,%d,%d\n",
+	   p2p_log.request, p2p_log.source, p2p_log.tag, p2p_log.count, p2p_log.comm);
+  }
+  return 0;
+}
+ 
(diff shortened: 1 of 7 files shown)


src/plugin/ipc/file/fileconnection.cpp
```diff
@@ -355,8 +355,11 @@ FileConnection::refill(bool isRestart)
         if (statbuf.st_size > _st_size &&
             ((_fcntlFlags & O_WRONLY) || (_fcntlFlags & O_RDWR))) {
           errno = 0;
-          JASSERT(truncate(_path.c_str(), _st_size) == 0)
-            (_path.c_str()) (_st_size) (JASSERT_ERRNO);
+          JWARNING(false) (_path) (_st_size) (statbuf.st_size)
+          .Text("Set saved size to the current file size");
```
gc00, line 359: `Set saved size` -> `Setting saved size`

contrib/mpi-proxy-split/mpi-wrappers/p2p-deterministic.c
```diff
@@ -54,35 +54,38 @@ void initialize_next_msg(int fd) {
...
   return (next_msg_entry.comm != MPI_COMM_NULL);
 }
 
-void get_next_msg(struct p2p_log_msg *p2p_msg) {
-  static int fd = -2;
-  if (fd == -2) {
+int get_next_msg(struct p2p_log_msg *p2p_msg) {
+  static int fd = -1;
+  int rc = 1;
+  if (fd == -1) {
```
gc00, line 72: Please revert to using `-2` (or `UNINITIALIZED`) here.  If I were more careful, I would have written:
```
#define UNINITIALIZED -2
static int fd = UNINITIALIZED;
...
if (fd == UNINITALIZED) {
```

There is a big difference between `UNINITIALIZED` and `-1` (meaning that some syscall had failed.

contrib/mpi-proxy-split/mpi-wrappers/p2p-deterministic.c
```diff
@@ -54,35 +54,38 @@ void initialize_next_msg(int fd) {
...
     char buf[100];
     int rank;
     MPI_Comm_rank(MPI_COMM_WORLD, &rank);
     snprintf(buf, sizeof(buf)-1, P2P_LOG_MSG, rank);
     fd = open(buf, O_RDONLY);
     if (fd == -1) {
-      perror("get_next_msg: open");
-      exit(1);
+      return 1;
```
gc00, line 79: Could you add a comment here?
Normally, if we can't open the file, I would expect to give up.  But if we think we can continue, shouldn't we at least return `-1`, so that the caller can take some action?

contrib/mpi-proxy-split/mpi-wrappers/mana_p2p_update_logs.c
```diff
@@ -66,16 +66,24 @@ void fill_in_log(struct p2p_log_msg *p2p_log) {
...
+    req_start_offset = 0;
+  }
   while (1) {
     readall(fd2, &p2p_request, sizeof(p2p_request));
     if (p2p_request.request == p2p_log->request) {
       p2p_log->source = p2p_request.source;
       p2p_log->tag = p2p_request.tag;
       break;
     }
+    // The request sequence is out of order
```
gc00, line 83: Could you add more detail to this comment?  What is an example scenario in which the request sequence can become out of order?
Thanks.

contrib/mpi-proxy-split/mpi-wrappers/p2p-deterministic.txt
```diff
@@ -142,7 +142,7 @@ ssize_t writeall(int fd, void *buf, size_t count) {
 // FIXME:  Need to add MPI_Improbe when MANA supports it
 #define LOG_PRE_Iprobe(status)
   MPI_Status local_status;
-  if (getenv("MANA_P2P_LOG") && status != MPI_STATUS_IGNORE) {
+  if (getenv("MANA_P2P_LOG") && status == MPI_STATUS_IGNORE) {
```
gc00, line 145: Nice catch!  Thanks for this bug fix.

gc00 (approved):
I am pre-approving in advance.
But please look carefully at my comments.  There is one important small change (keep using `UNINITIALIZED`, as different from `-1`, and there are 2 or 3 requests for more detailed comments.

contrib/mpi-proxy-split/mpi-wrappers/mana_p2p_update_logs.c
dahongli: I don't know what scenario causes out of order in the VASP test. I think this could happen with the following scenario. Since we save request info in MPI Wait, the request will be out of order in which the log file has request 2 first fallowed by request 1. 
T1: Irecv request 1
T2: Irecv request 2
T3: Wait request 2
T4: Wait request 1

contrib/mpi-proxy-split/mpi-wrappers/p2p-deterministic.c
dahongli: Yes. The first time we open this file is from REPLAY_PRE_IRECV(). At that time, this file has not created by LOG_POST_IRECV() yet. So the file does not exist and we expect it return "-1". We return `1` to tell caller there is no next message to replay. The caller will continue as if there is no next message to replay.

PR #108, opened 2022-04-27
Fix the race between two phase commit and checkpointing

dahongli:
There is race between two phase algorithm preSuspendBarrier() and two
phase commit commit_begin().
T1 thread #1: setCkptPending()
T2               thread #2: checks isCkptPending() then stop(comm)
T3              thread #2: stop(comm) sets state to PHASE_1
T4              thread #2: while isCkptPending() loop
T5 thread #1 calls waitForNewStateAfter(PHASE_1)

This change adds a timeout when waiting state change. When timeout
expires, the barrier sends a response message to the coordinator.
The coordinator gives a free phase1 pass to one member if all members
are in phase1 so this member can move to IN_CS state and contine
two phase commit. Once this member is in IN_CS, the coordinator would
unblock other members.

### DIFF
--- a/contrib/mpi-proxy-split/two-phase-algo.cpp
+++ b/contrib/mpi-proxy-split/two-phase-algo.cpp
@@ -217,7 +217,7 @@ TwoPhaseAlgo::preSuspendBarrier(const void *data)
   JASSERT(data).Text("Pre-suspend barrier called with NULL data!");
   DmtcpMessage msg(DMT_PRE_SUSPEND_RESPONSE);
 
-  phase_t st = getCurrState();
+  phase_t st;
   query_t query = *(query_t*)data;
   static int procRank = -1;
 
@@ -230,8 +230,10 @@ TwoPhaseAlgo::preSuspendBarrier(const void *data)
     case INTENT:
       setCkptPending();
       if (getCurrState() == PHASE_1) {
-        // If in PHASE_1, wait for us to finish doing IN_CS
-        if (waitForNewStateAfter(PHASE_1) != IN_CS) {
+	st = waitForNewStateAfter(PHASE_1, 10 /* timeout ms*/);
+	if (st == PHASE_1) break;
+        // Wait for us to finish doing IN_CS
+        if (st != IN_CS) {
           break;
         } else {
           while (waitForNewStateAfter(IN_CS) != IN_CS);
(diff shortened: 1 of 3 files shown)


gc00:
@dahongli ,  I'll try to look at this by tomorrow.

@xuyao0127 ,, I think you said you're quite busy right now.  If so, I'll take responsibility here.

gc00:
@dahongli , I'm still looking at this, and got distracted by some e-mail.  Could you start the larger testing of VASP6 even while I look at this?

Does it work well for you, if you combine this PR with PR #109 (by @jungan) in your own private branch?  Then you can continue testing VASP6 in your private branch, to see what other bugs might exist.  But I'm open to other strategies for testing VASP6 that you may have.

It's best if you can test your private branch on top of feature/dmtcp-master.  (That requires porting this PR to feature/dmtcp-master.  But if that isn't _easily_ feasible yet, then please continue testing here in the 'master' branch.

dahongli:
@gc00 I am testing my private branch on feature/dmtcp-master. Once I reproduced the race and verified the fix. I will send out a new PR against dmtcp-master branch. Next, I could try larger VASP6 testing.

contrib/mpi-proxy-split/two-phase-algo.cpp
```diff
@@ -230,8 +230,10 @@ TwoPhaseAlgo::preSuspendBarrier(const void *data)
     case INTENT:
       setCkptPending();
       if (getCurrState() == PHASE_1) {
-        // If in PHASE_1, wait for us to finish doing IN_CS
-        if (waitForNewStateAfter(PHASE_1) != IN_CS) {
+	st = waitForNewStateAfter(PHASE_1, 10 /* timeout ms*/);
```
gc00, line 233: Since this is a rare occurrence, let's set the timeout to 100 ms.  I'm worried that a timeout of 10 ms might add an extra 2% or 3% runtime overhead in some cases.

contrib/mpi-proxy-split/two-phase-algo.cpp
```diff
@@ -309,16 +311,27 @@ TwoPhaseAlgo::stop(MPI_Comm comm)
...
 
 phase_t
-TwoPhaseAlgo::waitForNewStateAfter(phase_t oldState)
+TwoPhaseAlgo::waitForNewStateAfter(phase_t oldState, int timeout_ms)
 {
   lock_t lock(_phaseMutex);
-  // The user thread will notify us if transition to any of these states
-  _phaseCv.wait(lock, [this, oldState]
+
+  if (timeout_ms) {
```
gc00, line 318: Suggested comment:
```
// Since _phaseCv is a condition var., it should be prepared for spurious wakeups.
// So, waking it up on a timeout is harmless.
```

gc00 (approved):
I'm approving this in advance.  Please note the suggested added comment, and using a 100 ms timeout to reduce runtime overhead.

PR #109, opened 2022-04-28
[MANA-113] Reserve 4GB more before libsStart

jungan:
In case the process calls `mmap`, there could have a VMA before the first VMA that maps the first library without any label. Adding 4G to avoid the issue


dahongli (approved):
Thanks Jun.

fyshhh:
wouldn't this reserve 8GB more?

PR #59, opened 2022-02-18
[MANA-87] Implement function wrappers Nimrod requires

fyshhh:
This pull request implements a few functions that nimrod requires:
- `MPI_Comm_get_attr`
- `MPI_Comm_set_attr`
- `MPI_File_open`
- `MPI_File_close`
- `MPI_File_get_size`
- `MPI_File_read_at`
- `MPI_Waitany`
- `MPI_Get_elements_x`
- `MPI_Type_size_x`

And also implements incidental functions that are either closely related or will likely be used (for debugging):
- `MPI_Comm_delete_attr`
- `MPI_Error_class`
- `MPI_Error_string`
- `MPI_Get_elements`

It also implements an upper-half only method of storing attributes. We use the following structure:
`[ attribute key -> { [communicator -> attribute value], copy callback function, delete callback function, extra state }]`
We reimplement two corresponding functions `MPI_Comm_create_keyval` and `MPI_Comm_free_keyval`, and add callback function calls to `MPI_Comm_free`.

This pull request also adds test cases for noteworthy function wrappers added.

PR #114, opened 2022-04-30
[MANA-87] Implement function wrappers Nimrod requires for feature/dmtcp-master

fyshhh:
This pull request implements a few functions that nimrod requires:
- `MPI_Comm_get_attr`
- `MPI_Comm_set_attr`
- `MPI_File_open`
- `MPI_File_close`
- `MPI_File_get_size`
- `MPI_File_read_at`
- `MPI_Waitany`
- `MPI_Get_elements_x`
- `MPI_Type_size_x`

And also implements incidental functions that are either closely related or will likely be used (for debugging):
- `MPI_Comm_delete_attr`
- `MPI_Error_class`
- `MPI_Error_string`
- `MPI_Get_elements`

It also implements an upper-half only method of storing attributes. We use the following structure:
`[ attribute key -> { [communicator -> attribute value], copy callback function, delete callback function, extra state }]`
We reimplement two corresponding functions `MPI_Comm_create_keyval` and `MPI_Comm_free_keyval`, and add callback function calls to `MPI_Comm_free`.

This pull request also adds test cases for noteworthy function wrappers added.

This is an updated version of #59, since we plan to support `feature/dmtcp-master`.


mpi-proxy-split/mpi-wrappers/mpi_error_wrappers.cpp
```diff
@@ -0,0 +1,60 @@
...
+USER_DEFINED_WRAPPER(int, Error_class, (int) errorcode, (int *) errorclass)
+{
+  int retval;
+  DMTCP_PLUGIN_DISABLE_CKPT();
+  JUMP_TO_LOWER_HALF(lh_info.fsaddr);
+  retval = NEXT_FUNC(Error_class)(errorcode, errorclass);
+  RETURN_TO_UPPER_HALF();
+  DMTCP_PLUGIN_ENABLE_CKPT();
+  return retval;
+}
```
jungan, line 43: Why not just use `DEFINE_FNC` ?

mpi-proxy-split/mpi-wrappers/mpi_file_wrappers.cpp
```diff
@@ -0,0 +1,91 @@
...
+  DMTCP_PLUGIN_DISABLE_CKPT();
+  JUMP_TO_LOWER_HALF(lh_info.fsaddr);
+  retval = NEXT_FUNC(File_get_size)(fh, size);
+  RETURN_TO_UPPER_HALF();
+  DMTCP_PLUGIN_ENABLE_CKPT();
+  return retval;
+
+}
+
+USER_DEFINED_WRAPPER(int, File_read_at, (MPI_File) fh, (MPI_Offset) offset,
```
jungan, line 72: ditto

mpi-proxy-split/mpi-wrappers/mpi_request_wrappers.cpp
```diff
@@ -169,6 +202,43 @@ USER_DEFINED_WRAPPER(int, Waitall, (int) count,
...
+          clearPendingRequestFromLog(array_of_requests[i]);
+          REMOVE_OLD_REQUEST(array_of_requests[i]);
+          array_of_requests[i] = MPI_REQUEST_NULL;
+        }
+        *index = i;
+        return retval;
+      }
+    }
+    if (all_null) {
+      return retval;
```
jungan, line 236: Could be potentially uninitialized

mpi-proxy-split/mpi-wrappers/mpi_request_wrappers.cpp
```diff
@@ -169,6 +202,43 @@ USER_DEFINED_WRAPPER(int, Waitall, (int) count,
...
+    }
+    sleep(1);
```
jungan, line 238: Why need it?

mpi-proxy-split/mpi-wrappers/mpi_request_wrappers.cpp
```diff
@@ -231,9 +301,8 @@ USER_DEFINED_WRAPPER(int, Probe, (int) source, (int) tag,
   return retval;
 }
 
-USER_DEFINED_WRAPPER(int, Iprobe,
-                     (int) source, (int) tag, (MPI_Comm) comm, (int*) flag,
-                     (MPI_Status *) status)
+USER_DEFINED_WRAPPER(int, Iprobe, (int) source, (int) tag, (MPI_Comm) comm,
+                     (int *) flag, (MPI_Status *) status)
```
jungan, line 305: What changed here?

mpi-proxy-split/mpi-wrappers/mpi_request_wrappers.cpp
```diff
@@ -249,8 +318,32 @@ USER_DEFINED_WRAPPER(int, Iprobe,
...
+                     (MPI_Datatype) datatype, (MPI_Count *) count)
+{
+  int retval;
+  DMTCP_PLUGIN_DISABLE_CKPT();
+  JUMP_TO_LOWER_HALF(lh_info.fsaddr);
+  retval = NEXT_FUNC(Get_elements_x)(status, datatype, count);
+  RETURN_TO_UPPER_HALF();
+  DMTCP_PLUGIN_ENABLE_CKPT();
+  return retval;
+}
```
jungan, line 343: Use `DEFINE_FNC` if possible, thanks!

jungan (changes requested):
After updated your code, please re-test it, and update the test result in https://memverge.atlassian.net/browse/MANA-108

As long as there is no blocking issue, we can merge it in first, and track the "crash when had more then 22 ranks" issue in a new ticket.

mpi-proxy-split/mpi-wrappers/mpi_comm_wrappers.cpp
```diff
@@ -235,44 +436,55 @@ USER_DEFINED_WRAPPER(int, Attr_put, (MPI_Comm) comm,
...
+// be resized.
+  if (keyvalVec.capacity() != MAX_KEYVAL_COUNT) {
+    keyvalVec.reserve(MAX_KEYVAL_COUNT);
   }
-  DMTCP_PLUGIN_ENABLE_CKPT();
+
+  int keyval = 0;
+  while (keyval < keyvalVec.size() + 1) {
+    if (keyval == keyvalVec.size()) {
+      keyvalVec.push_back(keyval);
```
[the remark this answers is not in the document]
fyshhh: Yes. I'll add a warning for it as you suggested below.

mpi-proxy-split/mpi-wrappers/mpi_comm_wrappers.cpp
```diff
@@ -235,44 +436,55 @@ USER_DEFINED_WRAPPER(int, Attr_put, (MPI_Comm) comm,
...
-  retval = NEXT_FUNC(Comm_free_keyval)(&realCommKeyval);
-  RETURN_TO_UPPER_HALF();
-  if (retval == MPI_SUCCESS && MPI_LOGGING()) {
-    // NOTE: We cannot remove the old comm_keyval from the map, since
-    // we'll need to replay this call to reconstruct any other comms that
-    // might have been created using this comm_keyval.
-    LOG_CALL(restoreComms, Comm_free_keyval, *comm_keyval);
+  int retval = MPI_SUCCESS;
+  int keyval = *comm_keyval;
+  if (keyval < keyvalVec.size()) {
```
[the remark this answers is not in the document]
fyshhh: Actually, the more I think about it, the more I feel that it should be harmless even if the vector is resized.

mpi-proxy-split/mpi-wrappers/mpi_request_wrappers.cpp
fyshhh: I was spacing out the `MPI_Test_internal` calls while testing so the logs wouldn't be cluttered. I can remove it if it's unnecessary.

mpi-proxy-split/mpi-wrappers/mpi_request_wrappers.cpp
fyshhh: I changed the formatting to be consistent with the other functions.

mpi-proxy-split/mpi-wrappers/mpi_comm_wrappers.cpp
```diff
@@ -436,53 +418,44 @@ USER_DEFINED_WRAPPER(int, Attr_put, (MPI_Comm) comm,
...
 
 USER_DEFINED_WRAPPER(int, Comm_free_keyval, (int *) comm_keyval)
 {
   int retval = MPI_SUCCESS;
   int keyval = *comm_keyval;
-  if (keyval < keyvalVec.size()) {
-// Assumes that it's the first time a comm_keyval is freed.
-    keyvalVec.at(keyval) = -1;
+  if (keyval < keyvalVec.size() && keyvalVec[keyval] != -1) {
+    keyvalVec[keyval] = -1;
```
jungan, line 458: What if `keyval >= keyvalVec.size()`, we at least should add log IMO.

mpi-proxy-split/mpi-wrappers/mpi_request_wrappers.cpp
jungan: I'm not sure how much impact it can have on performance, sleeping 1 second sounds a lot to me, and we shouldn't do it if it's unnecessary.

If you just want a better log, one option is to use it only for debug build, we shouldn't have it for release build.

mpi-proxy-split/mpi-wrappers/mpi_comm_wrappers.cpp
```diff
@@ -457,6 +457,8 @@ USER_DEFINED_WRAPPER(int, Comm_free_keyval, (int *) comm_keyval)
   if (keyval < keyvalVec.size() && keyvalVec[keyval] != -1) {
     keyvalVec[keyval] = -1;
     tupleMap.erase(keyval);
+  } else {
+    JWARNING(false).Text("Attempted to free an invalid key!");
```
jungan, line 455: It would help to print the value of `keyval` I guess

jungan (approved):
LGTM

Please give it a final test before merging it, and update the test result in the ticket, thanks!

jungan:
Could you please also re-write the commit history to make it more meaningful?

Try to combine them into several independent commits.

mpi-proxy-split/mpi-wrappers/mpi_comm_wrappers.cpp
```diff
@@ -122,6 +178,28 @@ MPI_Comm_free_internal(MPI_Comm *comm)
...
+    while (attributeMapIter != attributeMap->end()) {
+      MPI_Comm currComm = attributeMapIter->first;
+      if (currComm == *comm && tuple->_deleteFn != MPI_COMM_NULL_DELETE_FN) {
+        tuple->_deleteFn(*comm,
+                         tuplePair.first,
+                         attributeMap->at(currComm),
+                         tuple->_extraState);
+        attributeMapIter = attributeMap->erase(attributeMapIter);
+        break;
+      }
```
jungan, line 199: 1. Why not use `attributeMap->find(*comm)`?
2. If we found the `*comm` in the map, and the `_deleteFn` is null, should we remove it from the map?

mpi-proxy-split/mpi-wrappers/mpi_type_wrappers.cpp
```diff
@@ -157,6 +157,17 @@ USER_DEFINED_WRAPPER(int, Type_create_struct, (int) count,
...
+USER_DEFINED_WRAPPER(int, Type_size_x, (MPI_Datatype) type, (MPI_Count *) size)
+{
+  int retval;
+  DMTCP_PLUGIN_DISABLE_CKPT();
+  JUMP_TO_LOWER_HALF(lh_info.fsaddr);
+  retval = NEXT_FUNC(Type_size_x)(type, size);
+  RETURN_TO_UPPER_HALF();
+  DMTCP_PLUGIN_ENABLE_CKPT();
+  return retval;
+}
```
jungan, line 169: Why not `DEFINE_FNC` ?

mpi-proxy-split/mpi-wrappers/mpi_comm_wrappers.cpp
fyshhh: 1. will update it.
2. yes - in fact, we should remove it even if `_deleteFn` is not null. `_deleteFn` just does some clean-up with its arguments, but doesn't actually delete the communicator from the mapping.

thanks for catching these.

PR #123, opened 2022-05-04
Use the status MPI_SOURCE as rank in drain send-recv

dahongli:
The request saved in g_async_call could be MPI_ANY_SOURCE. When we calculate the bytes received from remote rank, we should use the actual rank we receive message from. This change gets the actual rank from the status MPI_SOURCE field.


mpi-proxy-split/unit-test/drain-send-recv-test.cpp
```diff
@@ -176,7 +176,13 @@ TEST_F(DrainTests, testRecvDrain)
...
   EXPECT_EQ(MPI_Type_size(MPI_INT, &size), MPI_SUCCESS);
   for (int i = 0; i < TWO; i++) {
-    EXPECT_EQ(MPI_Irecv(&rbuf, 1, MPI_INT, 0, 0, newcomm, &reqs[i]),
+    int source = 0;
+    int tag = 0;
+    if (i == 0) {
+      source = MPI_ANY_SOURCE;
+      tag = MPI_ANY_TAG;
+    }
+    EXPECT_EQ(MPI_Irecv(&rbuf, 1, MPI_INT, source, tag, newcomm, &reqs[i]),
```
jungan, line 185: I recall that we had a problem running this `drain-send-recv-test`, that's why we removed it from `TESTS` in `Makefile` @rohgarg may know the details.

jungan (approved):
Hi @dahongli , how about reverting the changes in the unit test(since it won't work) and merging this PR if it fixes MANA-122&MANA-123?
They block many test cases, I hope we can have a new build ASAP for Tom to verify it and continue the test cases.

PR #125, opened 2022-05-06
Add temporary fix for Fortran-to-C array bug

fyshhh:
This pull request adds a temporary fix for the Fortran-to-C array bug. This is a bug where, when passing an array as an argument in a function from Fortran to MANA's C wrapper, the arguments are passed by registers instead of on the stack. As a result, when another function returns inside the C wrapper, the arguments are corrupted. We don't yet know the cause of this bug, but creating a local copy of all the arguments before any internal function calls return fixes the issue.

### DIFF
--- a/mpi-proxy-split/mpi-wrappers/generate-mpi-fortran-wrappers.py
+++ b/mpi-proxy-split/mpi-wrappers/generate-mpi-fortran-wrappers.py
@@ -90,13 +90,17 @@ def emit_wrapper(decl, ret_type, fnc, args, arg_vars):
   if fnc in ['MPI_Wtime', 'MPI_Wtick']:
     print("  return " + fnc + "(" + cargs + ");")
   else:
+      if "int* index" in fargs:
+        # This is a temporary fix for the Fortran-to-C interface bug -
+        # more details can be found in MPI_Testany in mpi_request_wrappers.cpp
+        print("  int *local_index = index;");
       print("  *ierr = " + fnc + "(" + cargs + ");")
       if "int* index" in fargs:
         if not (fnc in ['MPI_Waitany', 'MPI_Testany']):
           warnings.warn("'int* index' found in {fnc} - likely to cause erroneous behavior".format(fnc=fnc))
         # We need this line because Fortran arrays start from 1, and *index
         # refers to the index of the first request that completes in an array.
-        print("  *index = *index + 1;")
+        print("  *local_index = *local_index + 1;")
       print("  return *ierr;")
   print("}")
   # print(ret_type + " " + fnc.lower() + "_ (" + args + ") __attribute__ ((weak, alias (\"" + fnc + "\")));")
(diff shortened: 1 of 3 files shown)


mpi-proxy-split/mpi-wrappers/generate-mpi-fortran-wrappers.py
```diff
@@ -90,13 +90,17 @@ def emit_wrapper(decl, ret_type, fnc, args, arg_vars):
   if fnc in ['MPI_Wtime', 'MPI_Wtick']:
     print("  return " + fnc + "(" + cargs + ");")
   else:
+      if "int* index" in fargs:
```
jungan, line 93: Is this reliable? Does this always mean the "index" of an array?

mpi-proxy-split/mpi-wrappers/generate-mpi-fortran-wrappers.py
```diff
@@ -90,13 +90,17 @@ def emit_wrapper(decl, ret_type, fnc, args, arg_vars):
...
+        # more details can be found in MPI_Testany in mpi_request_wrappers.cpp
+        print("  int *local_index = index;");
       print("  *ierr = " + fnc + "(" + cargs + ");")
       if "int* index" in fargs:
         if not (fnc in ['MPI_Waitany', 'MPI_Testany']):
           warnings.warn("'int* index' found in {fnc} - likely to cause erroneous behavior".format(fnc=fnc))
         # We need this line because Fortran arrays start from 1, and *index
         # refers to the index of the first request that completes in an array.
-        print("  *index = *index + 1;")
+        print("  *local_index = *local_index + 1;")
```
jungan, line 103: I'm confused why it makes a difference.

Unlike `Testall`&`Watiall` below, although you define a local variable `local_index` here, you don't pass the address of this local variable to the real MPI function. You still use the same "memory address" before&after, Why it can help?

mpi-proxy-split/mpi-wrappers/generate-mpi-fortran-wrappers.py
fyshhh: From what we've seen so far, yes - only `MPI_Testany` and `MPI_Waitany` use `int* index` as an argument in the wrappers we have implemented.

mpi-proxy-split/mpi-wrappers/generate-mpi-fortran-wrappers.py
fyshhh: The function originally looks like this:
```
EXTERN int mpi_testany_(..., int *index, ...) {
  *ierr = MPI_Testany(..., int *index, ...);
  *index = *index + 1;
  return *ierr;
}
```
But the bug affects the addresses passed into the functions, so after another function inside the `MPI_Testany` call returns, all references to `index` get jumbled up. We already have a fix for this within the `MPI_Testany` C wrapper, but because we have to do an operation on `index` outside of the C wrapper (we only increment `index` if Fortran calls it), we need a local copy of `index` that we operate on after the C wrapper returns.

jungan (approved):

PR #132, opened 2022-05-12
[MANA-133] Fix MPI_REQUEST_NULL not being skipped in Testany

fyshhh:
This pull requests fixes a bug where `MPI_REQUEST_NULL` handles were not being skipped properly in `MPI_Testany`.

### DIFF
--- a/mpi-proxy-split/mpi-wrappers/mpi_request_wrappers.cpp
+++ b/mpi-proxy-split/mpi-wrappers/mpi_request_wrappers.cpp
@@ -173,7 +173,6 @@ USER_DEFINED_WRAPPER(int, Testany, (int) count,
     if (local_array_of_requests[i] == MPI_REQUEST_NULL) {
       continue;
     }
-    *local_flag = 0;
     retval = MPI_Test(&local_array_of_requests[i], local_flag, local_status);
     if (retval != MPI_SUCCESS) {
       break;
--- a/mpi-proxy-split/mpi-wrappers/mpi_request_wrappers.cpp
+++ b/mpi-proxy-split/mpi-wrappers/mpi_request_wrappers.cpp
@@ -167,14 +167,13 @@ USER_DEFINED_WRAPPER(int, Testany, (int) count,
   MPI_Status *local_status = status;
 
   int retval;
-  bool all_requests_null = true;
-  *local_flag = 0;
+  *local_flag = 1;
   *local_index = MPI_UNDEFINED;
   for (int i = 0; i < local_count; i++) {
     if (local_array_of_requests[i] == MPI_REQUEST_NULL) {
       continue;
     }
-    all_requests_null = false;
+    *local_flag = 0;
     retval = MPI_Test(&local_array_of_requests[i], local_flag, local_status);
     if (retval != MPI_SUCCESS) {
       break;
@@ -183,9 +182,6 @@ USER_DEFINED_WRAPPER(int, Testany, (int) count,
       *local_index = i;
       break;
     }
-  }
-  if (all_requests_null) {
-    *local_flag = 1;
   }
   return retval;
 }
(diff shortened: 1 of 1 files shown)


mpi-proxy-split/mpi-wrappers/mpi_request_wrappers.cpp
```diff
@@ -156,9 +156,14 @@ USER_DEFINED_WRAPPER(int, Testany, (int) count,
   MPI_Status *local_status = status;
 
   int retval;
+  bool all_requests_null = true;
```
gc00, line 159: I think there is a simpler logic that doesn't need the variable `all_requests_null`.  I propose removing this variable, and making the changes noted below.

But if we must change the name, then this is better:
  `all_requests_null` -> `all_requests_are_null`

mpi-proxy-split/mpi-wrappers/mpi_request_wrappers.cpp
```diff
@@ -156,9 +156,14 @@ USER_DEFINED_WRAPPER(int, Testany, (int) count,
...
   *local_flag = 0;
```
gc00, line 160: Instead, let's initialize it as:  `*local_flag = 1;`

mpi-proxy-split/mpi-wrappers/mpi_request_wrappers.cpp
```diff
@@ -156,9 +156,14 @@ USER_DEFINED_WRAPPER(int, Testany, (int) count,
...
   *local_index = MPI_UNDEFINED;
   for (int i = 0; i < local_count; i++) {
+    if (local_array_of_requests[i] == MPI_REQUEST_NULL) {
+      continue;
+    }
+    all_requests_null = false;
```
gc00, line 166: Delete the line above.  We don't need it now.

mpi-proxy-split/mpi-wrappers/mpi_request_wrappers.cpp
```diff
@@ -168,6 +173,9 @@ USER_DEFINED_WRAPPER(int, Testany, (int) count,
       break;
     }
   }
```
gc00, line 175: Delete the next three lines:  `if (all_requests_null) { ... }`.

If `MPI_Test` was previously called and failed, then `MPI_Test` will have already set `*local_flag = 0;`.
If `MPI_Test` was never called, then `*local_flag` still has its initial value of `1`, which is the correct value when `all_requests_are_null` is true.

gc00 (commented):
@illio, Nice catch.

However, in my comments, I'm suggesting a simpler logic for doing the same thing.  We don't need `all_requests_null`.  Instead, we can simply initialize `*local_flag` to `1`.  This does the right thing when `all_requests_null` is true.

Could you make these changes, and then I'll look at it again?  Thanks.

fyshhh:
@gc00 thanks for your suggestions. I've removed `all_requests_null` and initialised `*local_flag` to 1. The only difference is that I added `*local_flag = 0` before each `MPI_Test` call. That seemed to be the necessary piece to tie it together - it seems like `*local_flag` was not being set to 0 in some cases.

mpi-proxy-split/mpi-wrappers/mpi_request_wrappers.cpp
```diff
@@ -167,9 +167,13 @@ USER_DEFINED_WRAPPER(int, Testany, (int) count,
...
 
   int retval;
-  *local_flag = 0;
+  *local_flag = 1;
   *local_index = MPI_UNDEFINED;
   for (int i = 0; i < local_count; i++) {
+    if (local_array_of_requests[i] == MPI_REQUEST_NULL) {
+      continue;
+    }
+    *local_flag = 0;
```
gc00, line 176: In principle, we should not need the line above (setting `*local_flag = 0;`).  I'd feel better if you can remove this line, and run it again to see why it's needed.  Logically, this should not change the result, since `MPI_Test` must also set `*local_flag`.  And it's dangerous to commit code that we don't understand.

gc00 (approved):
Please see why `*local_flag = 0` is needed, before pushing this in.  If it is needed, I'm worried that it indicates memory corruption or some other underlying issue.

mpi-proxy-split/mpi-wrappers/mpi_request_wrappers.cpp
fyshhh: I re-ran the test and it works without this line - I probably made a mistake in the prior tests.

PR #135, opened 2022-05-13
Fix Makefiles to use MANA_ROOT, not DMTCP_ROOT

gc00:
Prior to this, after a fresh 'git clone', one had to do `make -j mana` twice in order to install the MANA files in the top-level bin and lib/dmtcp directories.  It was because they were installing to dmtcp/{bin,lib} on the first round instead of to {bin,lib} where they belonged.  The details follow below.

  [ A second bug was discovered in doing this, and will be addressed in a future PR.  The MANA README.md says to define `MANA_ROOT`.  The MANA Makefiles use a make variable, `MANA_ROOT`.  The make variables then get accidentally initialized to the environment variable value.  If `MANA_ROOT` refers to a previous MANA build (not the current one), then a bug happens in which MANA tries to install to the previous MANA build. We should fix the Makefiles to avoid using such an environment variable. ]

 * In commit 7389f9, contrib/mpi-proxy-split was changed to
    mpi-proxy-split, as part of reorganizing to set DMTCP as a submodule.
    But unfortunately, parts of MANA were then set to install to
    MANA_ROOT/dmtcp/{bin,lib} instead of MANA_ROOT/{bin,lib}.

 * Without the current fix, MANA is installing into
    MANA_ROOT/dmtcp/{bin,lib} and that is later copied to
    MANA_ROOT/{bin,lib}.  As a result, it was necessary to 'make mana'
    twice (once to install in MANA_ROOT/dmtcp/{bin,lib},
    and once to copy to MANA_ROOT/{bin,lib}.

 * With this fix, MANA now installs directly to MANA_ROOT/{bin,lib}
    without disturbing MANA_ROOT/dmtcp.

### DIFF
--- a/mpi-proxy-split/Makefile
+++ b/mpi-proxy-split/Makefile
@@ -37,9 +37,13 @@ LIBOBJS = mpi_plugin.o p2p_drain_send_recv.o p2p_log_replay.o \
 
 LIBPROXY = libproxy.a
 
+# Modify if your MANA_ROOT is located elsewhere.
+ifndef MANA_ROOT
+  MANA_ROOT=..
+endif
 # Modify if your DMTCP_ROOT is located elsewhere.
 ifndef DMTCP_ROOT
-  DMTCP_ROOT=../dmtcp
+  DMTCP_ROOT=${MANA_ROOT}/dmtcp
 endif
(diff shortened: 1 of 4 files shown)


karya0 (approved):
LGTM

PR #136, opened 2022-05-13
Add minor update to mpi-proxy-split/README.md

fyshhh:
This pull request makes a minor update to `mpi-proxy-split/README.md`.

### DIFF
--- a/mpi-proxy-split/README.md
+++ b/mpi-proxy-split/README.md
@@ -75,7 +75,6 @@ See `man mana` or `nroff -man MANA_ROOT_DIR/manpages/mana.1` for the MANA man pa
 
    ```bash
    $ echo 'export PATH=$PATH:$MANA_ROOT/bin' >> ~/.bashrc
-   $ source ~/.bashrc
    ```
  
    On Cori, we need to allocate resources to run a job. The command above can be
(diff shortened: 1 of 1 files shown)


mpi-proxy-split/README.md
```diff
@@ -67,6 +67,12 @@ See `man mana` or `nroff -man MANA_ROOT_DIR/manpages/mana.1` for the MANA man pa
...
  
+   ```bash
+   $ export PATH=$PATH:$MANA_ROOT/bin
+   ```
+   
+   Optionally, you can make this path persistent by doing:
+
    ```bash
    $ echo 'export PATH=$PATH:$MANA_ROOT/bin' >> ~/.bashrc
    $ source ~/.bashrc
```
gc00, line 78: We don't want `source ~/.bashrc` here.  The idea is that we added to our path in the current shell, and optionally, we can add to the path in .bashrc for the sake of future shells.  So, there is no reason to do `source ~/.bashrc` in the current shell.  We've already extended the path  We don't need to do it again for this shell.

gc00 (approved):
I'm approving the PR now.  But please delete that line `source ~/.bashrc`, as indicated in my comment.

PR #138, opened 2022-05-13
[MANA-139] Skip replaying the completed request

dahongli:
This change use unsorted map to quickly find if the request in the record-replay has been completed. If it is completed, then we will skip replay this record in checkpoint restart.


gc00 (approved):
LGTM.  Let's push it in now, before the nersc-alpha-2 tag.

I've sent mail to @xuyao0127 and @karya0 to take a look at this, and then to create the nersc-alpha-2 tag at the end of the weekend or on Monday.

PR #158, opened 2022-05-23
Define NO_BARRIER_BCAST

dahongli:
When NO_BARRIER_BCAST is defined, p2p is used to do broadcast. Currently this is a workaround to get the bit-for-bit match with native run for VASP checkpoint restart.

### DIFF
--- a/mpi-proxy-split/mpi-wrappers/mpi_collective_wrappers.cpp
+++ b/mpi-proxy-split/mpi-wrappers/mpi_collective_wrappers.cpp
@@ -50,10 +50,10 @@ isUsingCollectiveToP2p() {
 #endif
 }
 
-// #define NO_BARRIER_BCAST
 using namespace dmtcp_mpi;
 
 #ifndef MPI_COLLECTIVE_P2P
+#define NO_BARRIER_BCAST
 #ifdef NO_BARRIER_BCAST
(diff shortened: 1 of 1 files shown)


xuyao0127:
Why NO_BARROER_BCAST needs to be enabled?

dahongli:
> Why NO_BARROER_BCAST needs to be enabled?

It seems like there is some issue in handling the collective version Bcast at checkpoint restart. It is not clear what the root cause of it yet. When the collective Bcast is used, VASP5 restart consistently shows bit-for-bit issue compared with VASP5 native run. This bit-for-bit issue disappears with NO_BARRIER_BCAST enabled. 
While we look into the cause of this issue, we could enable this to get the correctness. Once we solve the underlying issue, we can disable it and use collective version Bcast.

xuyao0127 (approved):
This PR is only for debugging the bit-for-bit issue related to MPI_Bcast. We should use the collective wrapper after fixing the bit-for-bit issue.

dahongli:
> This PR is only for debugging the bit-for-bit issue related to MPI_Bcast. We should use the collective wrapper after fixing the bit-for-bit issue.

Totally agreed. I will change back to use collective wrapper once the underlying issue is resolved.

Issue #161, opened 2022-05-30
Deadlock in the pre-ckpt phase when the checkpoint interval is short


JainTwinkle:
Quoting @marcpb94 from #133:

> It seems to be working now! I tried with mpi_hello_world with 1 rank and a heat distribution application (that we usually use for testing) with 4 ranks.

> However, there seems to be an issue we already noticed in the old version of MANA, and we were hoping it was fixed in the new version. It seems to happen with our heat distribution application, I attach the source code so that you are able to reproduce the problem. [heatdis.zip](https://github.com/mpickpt/mana/files/8797781/heatdis.zip)

> The issue is that when checkpointing with relatively short intervals(and letting it execute for a few minutes), the execution eventually encounters a deadlock in the pre-checkpoint phase. Specifically, at least one of the MPI processes gets stuck in the drainSendRecv() function in the DMTCP_EVENT_PRECHECKPOINT case from mpi_plugin_event_hook().

JainTwinkle:
@marcpb94,

Could you provide the backtrace of the hanging process (all threads) and specific instructions to reproduce the issue? For example, the launch command, input to the executable, and short interval (2 sec?).

Thanks!

gc00:
@JainTwinkle  @marcpb94  Does this dreadlock occur if you choose a longer checkpoint interval, such as 5 minutes?
This might be a long-known issue for short checkpoint intervals.  We were not concentrating on it earlier, when we were concentrating on making MANA more robust.
But now that MANA is becoming more robust, this might be a good time to analyze what is causing the dreadlock.

JainTwinkle:
Thanks, @marcpb94! I was able to reproduce this issue. We are looking into it.

@karya0  @xuyao0127   @dahongli 
Rank 3 is the one that is currently stuck. Please find the backtrace of all four ranks here: [heat-distribution-4-ranks-backtrace.txt](https://github.com/mpickpt/mana/files/8808794/heat-distribution-4-ranks-backtrace.txt)

Yao,
We suspect that this might be an issue in the design while draining messages from `MPI_Isend`, and we think you would better know about the pre-ckpt phase algorithm. Could you please look at the backtraces?
The source is available here: [heatdis.zip](https://github.com/mpickpt/mana/files/8797781/heatdis.zip)

JainTwinkle:
Update: @xuyao0127 says that he is able to reproduce the issue on Cori, and he is working on it.

xuyao0127:
[...]

PR #165, opened 2022-06-07
Async p2p bugfix

xuyao0127:
This is a fix of issue #161. When sharing large data with asynchronous point-to-point communication, MPICH requires a handshake between the sender and receiver. If MPI_Irecv has not been called before checkpoint time, only the metadata of the message is shared among MPI processes and the data is still in the sender's local buffer. As a result, MPI_Iprobe in the p2p draining algorithm can detect the incoming message, but MPI_Irecv can't receive the message. In this PR, I also call MPI_Test on pending MPI_Isend requests to complete potential pending handshakes.

### DIFF
--- a/mpi-proxy-split/p2p_drain_send_recv.cpp
+++ b/mpi-proxy-split/p2p_drain_send_recv.cpp
@@ -97,8 +97,9 @@ recvMsgIntoInternalBuffer(MPI_Status status, MPI_Comm comm)
   MPI_Type_size(MPI_BYTE, &size);
   JASSERT(size == 1);
   void *buf = JALLOC_HELPER_MALLOC(count);
-  MPI_Recv(buf, count, MPI_BYTE, status.MPI_SOURCE, status.MPI_TAG,
+  int retval = MPI_Recv(buf, count, MPI_BYTE, status.MPI_SOURCE, status.MPI_TAG,
            comm, MPI_STATUS_IGNORE);
+  JASSERT(retval == MPI_SUCCESS);
 
   mpi_message_t *message = (mpi_message_t *)JALLOC_HELPER_MALLOC(sizeof(mpi_message_t));
   message->buf        = buf;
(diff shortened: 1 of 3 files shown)


mpi-proxy-split/p2p_drain_send_recv.cpp
```diff
@@ -115,48 +116,46 @@ recvMsgIntoInternalBuffer(MPI_Status status, MPI_Comm comm)
...
+    int flag = 0;
+    MPI_Status status;
+    MPI_Test_internal(&request, &flag, &status, false);
+    if (flag) {
+      if (call->type == IRECV_REQUEST) {
         int size = 0;
         MPI_Type_size(call->datatype, &size);
         int worldRank = localRankToGlobalRank(status.MPI_SOURCE,
-                                              call->comm);
+            call->comm);
```
gc00, line 136: This is minor, but I find that the code is more readable when this argument `call->comm` is indented under the first argument (under `status`), as it was before.

mpi-proxy-split/p2p_drain_send_recv.cpp
```diff
@@ -115,48 +116,46 @@ recvMsgIntoInternalBuffer(MPI_Status status, MPI_Comm comm)
...
     mpi_async_call_t *call = it->second;
-    if (call->type == IRECV_REQUEST) {
-      int flag = 0;
-      MPI_Status status;
-      MPI_Test_internal(&request, &flag, &status, false);
-      if (flag) {
-        UPDATE_REQUEST_MAP(request, MPI_REQUEST_NULL);
+    int flag = 0;
+    MPI_Status status;
+    MPI_Test_internal(&request, &flag, &status, false);
```
gc00, line 130: Can we add a comment above this similar to your comment in this PR?  This will help others to understand the code better.  You can either insert your entire PR comment, which is very well written, or if you're searching for something shorter, perhaps:
```
// This is needed if an MPI_Isend was called earlier.  Without this,
// large messages within the same node will fail under MPICH and some
// other MPIs.  A previous call to MPI_Irecv caused only the metadata to be
// exchanged.  So, MPI_Iprobe succeeds and MPI_Irecv will later fail, unless
// we force the sending of data via MPI_Test. 
```

mpi-proxy-split/p2p_drain_send_recv.cpp
```diff
@@ -213,11 +212,23 @@ bool
 allDrained()
 {
   int i;
+  // P2P counters are matched
   for (i = 0; i < g_world_size; i++) {
     if (g_bytesSentToUsByRank[i] > g_recvBytesByRank[i]) {
       return false;
     }
   }
+  // No pending MPI_Isend request
```
gc00, line 221: Let's have a more expressive comment.  Perhaps something like:
```
// recvMsgIntoInternalBuffer returns false if no MPI_Isend request is found:
```

mpi-proxy-split/p2p_drain_send_recv.cpp
```diff
@@ -228,8 +239,8 @@ drainSendRecv()
   while (!allDrained()) {
     sleep(1);
     int numReceived = 0;
-    // If pending MPI_Irecvs, use MPI_Test in case msg was sent.
-    numReceived += completePendingIrecvs();
+    // If pending MPI_Irecvs or MPI_Isend, use MPI_Test to try to complete them.
```
gc00, line 242: `pending MPI_Irecvs` -> `pending MPI_Irecv`
(I know this is a previous comment.  But let's polish it further now.)

`to complete them` -> `to complete it`  ("A or B" is singular, and "A and B" is plural.)

mpi-proxy-split/test/large_async_p2p.c
```diff
@@ -0,0 +1,48 @@
+#include <mpi.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <assert.h>
+#include <unistd.h>
+
+#define MSG_SIZE 64*1024 // large message
+#define MSG_SIZE 64 // small message
```
gc00, line 8:
```
#define MSG_SIZE 64*1024 // large message
#ifndef MSG_SIZE
# define MSG_SIZE 64 // small message
#endif
```

1. Some compilers will complain about re-defining MSG_SIZE without this.
The default version uses a large message in this version.  But I think this is a good thing, since that's what caused the original bug.
2. Should we raise the large message size to `256*1024`, to make it more future-proof?  In the future, maybe the optimized hardware considers even `64*1024` to be a small message that can use an EAGER policy.

mpi-proxy-split/test/large_async_p2p.c
```diff
@@ -0,0 +1,48 @@
+#include <mpi.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <assert.h>
+#include <unistd.h>
+
```
gc00, line 6: Let's add some comment like:
```
// For testing, checkpoint this during the 5-second sleep between
// MPI_Isend (rank 1) and MPI_Irecv (rank 0).
```

gc00 (approved):
LGTM.
I'm approving it in advance, but please see the minor suggestions (mostly comments) to add.

mpi-proxy-split/test/large_async_p2p.c
xuyao0127: This is a mistake. I was thinking of commenting out one of two definitions.

PR #169, opened 2022-06-15
Dynamically choose lh_proxy version for NIMROD

fyshhh:
This pull request adds two things:

1. building MANA now builds two copies of `lh_proxy`
    * `lh_proxy`, which will be loaded into `0xe000000`
    * `lh_proxy_default_address`, which will be loaded into `0x400000`
1. `mana_launch` now checks if the application is NIMROD, and if so will use `lh_proxy_default_address` instead

And the documentation has also been updated to reflect these changes, since manually building `lh_proxy` is no longer necessary.

### DIFF
--- a/bin/mana_launch
+++ b/bin/mana_launch
@@ -107,6 +107,15 @@ if [ "$coordinator_found" == 0 ]; then
   exit 5
 fi
 
+# At higher ranks, restarting NIMROD causes a heap overflow (for a more detailed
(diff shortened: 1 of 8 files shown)


JainTwinkle:
@fyshhh, let me know once you add support to choose the correct `lh_proxy` during restart. I'll review it afterward.

dahongli (approved):
Looked the latest commit "load lh_proxy dynamically at restart". It looks good.

PR #172, opened 2022-06-15
Update tests to run for longer time

chirag-singh-memverge:
I updated the tests to run for a predefined amount of wall time (including control over how long to sleep for)
I left out many of the send/recv tests because these methods are already covered thoroughly by the tests included


chirag-singh-memverge:
@JainTwinkle Please review

gc00:
@dahongli , This PR looks good, and I was about to approve it.  But you use the test s more often than I do.  Could you give this a quick look, and approve it if it seems okay?
We don't need a long and careful review.  If some test no longer runs well, it will be easy to individually fix that test later.

dahongli (changes requested):
Thanks Chirag for the change. Overall, it looks good except Ibcast_test.

dahongli (approved):
Looks good. Thanks Chirag!

PR #175, opened 2022-06-20
Updated DMTCP and do not skip libhugetlbfs.so.

karya0:
Libhugetlbfs.so shouldn't be skipped during checkpoint otherwise our linkmaps, etc., will point to invalid memory.


gc00 (approved):
LGTM.  This is definitely the right thing to do.  At the time, we had omitted restoring the libtlbfs.so in the mistaken belief that we were trying to compile DMTCP to not use hugepages.  It's true that we want to compile DMTCP to not use hugepages.  But the other memory segments need this user-space component in order to properly manage hugepages.  Maybe VASP5 is failing on restart because it tries to create a hugepage during restart.

PR #177, opened 2022-06-23
Feature/seq num

xuyao0127:
Implemented a hybrid two-phase-commit algorithm based on sequence numbers of communicators. When I was testing this PR on Cori, it can checkpoint and restart with small programs. But when restarting VASP, there's a segmentation fault accessing the TLS area.


xuyao0127:
Found an issue when running on multiple nodes.

xuyao0127:
Have tested the VASP 5 CaPOH test case with 128 ranks on 2 KNL nodes.

PR #181, opened 2022-06-24
Set collective version bcast as default

dahongli:
P2p version bcast was a workaround for correctness. Since the
collective version bcast correctness issue is resolved, this
change sets the collective version bcast as the default.

This changes solves the performance regression introduced by p2p bcast.


xuyao0127 (approved):
LGTM

PR #185, opened 2022-06-29
Added util to support debugging issues only arising at large numbers of ranks

chirag-singh-memverge:


util/attach_all.py
```diff
@@ -0,0 +1,86 @@
...
+import socket
+
+from subprocess import PIPE
+
+outlock = threading.Lock()
+
+
+def parse_ranks():
+    ranks = {}
+    stat = subprocess.run(['mana_status'], stdout=PIPE)
```
JainTwinkle, line 14: Let's add logic here to check if mana_status is not in the path or if someone wants to provide an absolute path to the mana_status script.

chirag-singh-memverge: Fixed

gc00 (approved):
I'm approving this in advance.  But I have two comments:
1.  In the first commit, you have a `print_usage` function.  I think that is quite usefful.  It seems to have disappeared in the second commit.
2. I hope you will squash the two commits into one, before doing the 'Rebase and Merge'.    (Since it's a new Python file, we don't need it to be split over two commits.)

And as before, please use the 'Rebase and Merge' -- not 'merge commits'.

chirag-singh-memverge:
>     1. In the first commit, you have a `print_usage` function.  I think that is quite usefful.  It seems to have disappeared in the second commit.

I had removed this because it clogged stdout when we want to see which rank was skipped for manual debugging. I have added it back with a debug flag that makes it optional.

chirag-singh-memverge:
> In the first commit, you have a print_usage function. I think that is quite usefful. It seems to have disappeared in the second commit.

It was removed because the argument parser will print a similar help message if it receives an invalid argument or the --help flag.

PR #191, opened 2022-07-06
Print error msg if MPI_Cart_create() is called more than once.

tarunsmalviya:
Current implementation only supports one cartesian communicator. So, print error msg if MPI_Cart_create() is called more than once.


xuyao0127 (approved):
LGTM

PR #194, opened 2022-07-08
Reverted to support for multiple cartesian coordinators

chirag-singh-memverge:
- Reverted changes to cartesian support that reordered ranks in cartesian-supporting communicators with new macro to enable these changes
- Changed Cart_create to always set reorder=false to ensure that restarted ranks have the same cartesian coordinates in all communicators

### DIFF
--- a/mpi-proxy-split/mpi-wrappers/mpi_cart_wrappers.cpp
+++ b/mpi-proxy-split/mpi-wrappers/mpi_cart_wrappers.cpp
@@ -30,19 +30,13 @@
 #include "mpi_nextfunc.h"
 #include "record-replay.h"
 #include "virtual-ids.h"
+#ifdef SINGLE_CART_REORDER
 #include "two-phase-algo.h"
+#include "../cartesian.h"
+#endif
 #include "p2p_drain_send_recv.h"
 
-#include "../cartesian.h"
-
 using namespace dmtcp_mpi;
-
-// This variable holds the cartesian properties and is only used at the time of
-// checkpoint (DMTCP_EVENT_PRECHECKPOINT event in mpi_plugin.cpp).
-CartesianProperties g_cartesian_properties = { .comm_old_size = -1,
-                                               .comm_cart_size = -1,
-                                               .comm_old_rank = -1,
-                                               .comm_cart_rank = -1 };
 
 USER_DEFINED_WRAPPER(int, Cart_coords, (MPI_Comm) comm, (int) rank,
                      (int) maxdims, (int*) coords)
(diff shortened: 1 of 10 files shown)


restart_plugin/mtcp_restart_plugin.c
```diff
@@ -420,7 +423,7 @@ mtcp_plugin_hook(RestoreInfo *rinfo)
   // Refer to "blocked memory" in MANA Plugin Documentation for the addresses
   // FIXME:  Rewrite this logic more carefully.
   char *start1, *start2, *end1, *end2;
-  if (rinfo->maxLibsEnd + 1 * GB < rinfo->minHighMemStart /* end of stack of upper half */) {
+  if (rinfo->maxLibsEnd + 1 * GB < rinfo->minHighMemStart  end of stack of upper half ) {
```
JainTwinkle, line 426: you'll need the `/* .. */` for the comment.

mpi-proxy-split/unit-test/Makefile
```diff
@@ -53,7 +53,7 @@ default: ${TEST_BINS}
 ../record-replay.o ../drain_send_recv_packets.o:
 	@make -C ..
 
-%.exe: %.o ../record-replay.o ../split_process.o ../lower-half/procmapsutils.o
+%.exe: %.o ../record-replay.o #../split_process.o ../lower-half/procmapsutils.o
```
JainTwinkle, line 56: There's no harm in keeping these objects files. It's compatible with and without the `SINGLE_CART_REORDER` macro. Let's put a Note above this rule that these two objects are needed with the macro, and one should remove these two object files if they choose to remove the single_cart_reorder implementation.

mpi-proxy-split/unit-test/Makefile
```diff
@@ -53,7 +53,8 @@ default: ${TEST_BINS}
 ../record-replay.o ../drain_send_recv_packets.o:
 	@make -C ..
 
-%.exe: %.o ../record-replay.o ../split_process.o ../lower-half/procmapsutils.o
+# UNCOMMENT the objects below if compiling with SINGLE_CART_REORDER
+%.exe: %.o ../record-replay.o # ../split_process.o ../lower-half/procmapsutils.o
```
JainTwinkle, line 57: MANA won't be able to build when compiling with SINGLE_CART_REORDER. 
You need to keep them by default and add a comment that these object files are needed by this macro and they should be removed if we no longer use the single_cart_reorder implementation.

mpi-proxy-split/cartesian.h
```diff
@@ -1,6 +1,8 @@
 #ifndef CARTESIAN_H
 #define CARTESIAN_H
 
+#ifdef SINGLE_CART_REORDER
```
JainTwinkle, line 4: I don't think you need to comment everything out within this header file. You can take care of that while including this header file. It means to include this header file only when the macro is defined.

mpi-proxy-split/mpi-wrappers/mpi_cart_wrappers.cpp
```diff
@@ -222,6 +161,89 @@ USER_DEFINED_WRAPPER(int, Dims_create, (int)nnodes, (int)ndims, (int *)dims)
...
+  return twoPhaseCommit(old_comm, realBarrierCb);
+}
+#else
+
+USER_DEFINED_WRAPPER(int, Cart_create, (MPI_Comm) old_comm, (int) ndims,
+                     (const int*) dims, (const int*) periods, (int) reorder,
+                     (MPI_Comm *) comm_cart)
+{
+  int retval;
+  reorder = 0;
```
JainTwinkle, line 225: Let's add a comment above this explaining why we are enforcing reorder to be false.

mpi-proxy-split/mpi_plugin.cpp
```diff
@@ -386,9 +396,9 @@ mpi_plugin_event_hook(DmtcpEvent_t event, DmtcpEventData_t *data)
       dmtcp_local_barrier("MPI:Reset-Drain-Send-Recv-Counters");
       resetDrainCounters(); // p2p_drain_send_recv.cpp
       mana_state = RESTART_REPLAY;
-      dmtcp_global_barrier("MPI:setCartesianCommunicator");
+      //dmtcp_global_barrier("MPI:setCartesianCommunicator");
       // record-replay.cpp
-      setCartesianCommunicator(lh_info.getCartesianCommunicatorFptr);
+      //setCartesianCommunicator(lh_info.getCartesianCommunicatorFptr);
```
JainTwinkle, line 401: shouldn't it be under the `#ifdef .. #end` block?

mpi-proxy-split/mpi_plugin.cpp
```diff
@@ -46,12 +48,16 @@
 #include "jfilesystem.h"
 #include "protectedfds.h"
 #include "procselfmaps.h"
+#ifdef SINGLE_CART_REORDER
```
JainTwinkle, line 51: Minor: you can make it a single `#ifdef` block by putting `signal.h, sys/personality.h, sys/mman.h` on top of `stdio.h` and by putting `cartesian.h` right under `sys/stat.h`.

mpi-proxy-split/record-replay.cpp
```diff
@@ -20,11 +20,15 @@
  ****************************************************************************/
 
 #include <mpi.h>
+#ifdef SINGLE_CART_REORDER
 #include <fcntl.h>
+#endif
 #include "jassert.h"
 #include "jconvert.h"
 
+#ifdef SINGLE_CART_REORDER
```
JainTwinkle, line 29: Minor: you can make it a single `#ifdef` block by putting `cartesian.h` right under `fcntl.h`.

mpi-proxy-split/mpi-wrappers/mpi_cart_wrappers.cpp
```diff
@@ -222,6 +161,93 @@ USER_DEFINED_WRAPPER(int, Dims_create, (int)nnodes, (int)ndims, (int *)dims)
...
+CartesianProperties g_cartesian_properties = { .comm_old_size = -1,
+                                               .comm_cart_size = -1,
+                                               .comm_old_rank = -1,
+                                               .comm_cart_rank = -1 };
+
+USER_DEFINED_WRAPPER(int, Cart_create, (MPI_Comm)old_comm, (int)ndims,
+                     (const int *)dims, (const int *)periods, (int)reorder,
+                     (MPI_Comm *)comm_cart)
+{
+  //JASSERT(g_cartesian_properties.comm_old_size == -1)
```
JainTwinkle, line 176: Let's enable the check here and change `JASSERT` to `JWARNING`.

mpi-proxy-split/unit-test/Makefile
```diff
@@ -53,7 +53,9 @@ default: ${TEST_BINS}
 ../record-replay.o ../drain_send_recv_packets.o:
 	@make -C ..
 
-%.exe: %.o ../record-replay.o ../split_process.o ../lower-half/procmapsutils.o
+# COMMENT out the objects split_process.o and procmapsutils.o
```
JainTwinkle, line 56: Please change the comment to the following:
```
# NOTE:  The objects files split_process.o and procmapsutils.o
below are required only when MANA is configured with SINGLE_CART_REORDER C/C++ flag.
One should remove these two obj files when we decide to remove the SINGLE_CART_REORDER macro
implementation entirely from MANA. 
```

mpi-proxy-split/mpi-wrappers/mpi_cart_wrappers.cpp
```diff
@@ -222,6 +161,93 @@ USER_DEFINED_WRAPPER(int, Dims_create, (int)nnodes, (int)ndims, (int *)dims)
...
+
+USER_DEFINED_WRAPPER(int, Cart_create, (MPI_Comm) old_comm, (int) ndims,
+                     (const int*) dims, (const int*) periods, (int) reorder,
+                     (MPI_Comm *) comm_cart)
+{
+  int retval;
+  // We are forcing reordering to be false because in order to ensure that on
+  // restart, each rank has the same cartesian coordinates for each of
+  // communicator that it is part of. This may cause some runtime overhead
+  // on restart on applications that call Cart_create with reorder=true
```
JainTwinkle, line 228: Please change the comment like following:
```
The MPI library assigns the cartesian coordinates naively (in the increasing rank order) with no reordering as opposed to optimal reordering based on the physical topology.
For example, if there are 6 ranks (0-5) exist, then without reordering, the two-dimensional coordinates will be following:
(0, 0) -> rank 0
(0, 1) -> rank 1
.
.
(2, 2) -> rank 5
While no reordering ensures the same rank to coordinates mapping on the restart,  it can incur some overhead as MPI can no longer optimize the ranks' ordering. For now, we focus on the correctness across checkpoint-restart. Therefore, we enforce the reorder variable to be false.

FIXME: Handle the case when reorder variable is true.
```
and add a warning message whenever reorder is true:
```
JWARNING (reorder == false) .Text ("We are enforcing reorder to false as the current implementation does not support reordered ranks.");
```

JainTwinkle (approved):
Please address the comments. The rest looks okay to me.

Issue #195, opened 2022-07-11
Problem with the MPI_Type_hvector wrapper for iPic3D


Marc-Miranda:
Good afternoon,
Some days ago I asked whether a wrapper for MPI_Type_hvector could be implemented, as iPic3D required it. @chirag-singh-memverge was kind enough to implement it. Thank you! However, the assumption that the stride must be a multiple of the size of the type does not hold for iPic3D. Is there an easy fix to this?

Thank you,
Marc

chirag-singh-memverge:
This PR should have the fix: https://github.com/mpickpt/mana/pull/196

chirag-singh-memverge:
@Marc-Miranda Please apply the PR + test to see if the issue is resolved so we can merge

Marc-Miranda:
Hello!
Sorry for the delay. I did not have access to the computer during the last week. I have now checked it and it seems to work. Now I get a new error for a different implemented wrapper, but not this. I will open a new issue, so we can close this. Thank you both!

JainTwinkle:
Thanks for the confirmation!

PR #196, opened 2022-07-11
Refactored MPI_Type_vector to support MPI_Type_hvector

chirag-singh-memverge:
- Removed previous FIXME and refactored MPI_Type_vector to use MPI_Type_hvector
- Required changing the record/replay to use MPI_Type_hvector instead
- Added support in record/replay for long datatype (to support MPI_Aint parameter)
- Added test for MPI_Type_vector


gc00 (approved):
I updated the description to include a reference to fixing Issue #195, so people will know the motivation.

@Marc-Miranda, Could you test out this branch, `chirag-singh-memverge:Type_vector-refactoring`?

@chirag-singh-memverge, I'm approving this now.  Maybe wait until Tuesday to see if @Marc-Miranda has any comments, and if not, then go ahead and push this in then.

PR #198, opened 2022-07-15
Fix MANA-161: find nimrod as a substring

JainTwinkle:
Use `lh_proxy_da` when the executable has the keyword `nimrod` in its name. Earlier, the logic expects the executable's name to be exactly `nimrod`. It's causing nimrod's regression test to be failed as `mana_launch` would choose the wrong `lh_proxy`. The PR fixes that issue.

### DIFF
--- a/bin/mana_launch
+++ b/bin/mana_launch
@@ -112,7 +112,7 @@ fi
 # A temporary workaround is to swap the memory addresses of NIMROD and lh_proxy,
 # which requires a specially built version of lh_proxy.
 app_name="`echo $options | head -n1 | xargs basename`"
-if [ "$app_name" == "nimrod" ]; then
+if echo $app_name | grep -q nimrod ; then
   export USE_LH_PROXY_DEFADDR=1
 fi
 


bin/mana_launch
```diff
@@ -112,7 +112,7 @@ fi
 # A temporary workaround is to swap the memory addresses of NIMROD and lh_proxy,
 # which requires a specially built version of lh_proxy.
 app_name="`echo $options | head -n1 | xargs basename`"
-if [ "$app_name" == "nimrod" ]; then
+if [[ $app_name == *"nimrod"* ]]; then
```
gc00, line 115: @JainTwinkle,
    Thanks for catching this bug in MANA-161.  Great catch.  I have a suggested alternate fix.

I think you are using wild cards (`*`) to look for a file name with the substring 'nimrod'.  That depends on looking in the local directory, and might depend on whether the user's PATH includes `.`.

What about something like:
```
if echo $app_name | grep nimrod; then
```
I think that works, and it is less dangerous.

gc00 (commented):
Please see my inline comment.

JainTwinkle:
@gc00 The PR is updated as suggested. Could you please review and approve it?

bin/mana_launch
```diff
@@ -112,7 +112,7 @@ fi
 # A temporary workaround is to swap the memory addresses of NIMROD and lh_proxy,
 # which requires a specially built version of lh_proxy.
 app_name="`echo $options | head -n1 | xargs basename`"
-if [ "$app_name" == "nimrod" ]; then
+if ( echo $app_name | grep -q  nimrod ) then
   export USE_LH_PROXY_DEFADDR=1
```
gc00, line 116: Please remove `(` and `)`.  There's no reason to ton this as a subprocess.

gc00 (approved):

bin/mana_launch
gc00: @JainTwinkle , please change this line before committing.  I am approving in advance.

JainTwinkle: fixed and merged!

PR #199, opened 2022-07-15
Mark platform as Cori for a compute node

JainTwinkle:
Issue: On Cori, the `HOST` env. var. points to the login node (`coriXXX`) when one logs into the compute node via the `salloc` command. However, if you ssh explicitly to your allocated compute node via the `JOB_ID`, the `HOST` points to the compute node. At this point, MANA's build fails as the Makefile considers it as a non-cori platform.
The PR handles the case when one builds MANA on a compute node and the `HOST` points to the compute node.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -45,7 +45,7 @@ default: display-build-env add-git-hooks mana_prereqs
 
 all: default
 
-mana: dmtcp
+mana: mana_prereqs dmtcp
 	cd mpi-proxy-split && $(MAKE) install && $(MAKE) -j tests
 
 dmtcp:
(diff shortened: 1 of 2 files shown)


gc00 (commented):
@JainTwinkle ,
    I trust your knowledge on this.  But is it recommended to build and compile on nid0*?
    At NERSC, they recommend:
* Compile on login nodes to run on compute nodes : https://www.nersc.gov/assets/Uploads/06-Compiling-Codes.pdf
* The login nodes are defined as cori* : See https://docs.nersc.gov/systems/cori/#login-nodes

Maybe it would be better to catch if someone is configuring on nid0* and tell them to instead configure and build on cori*.
We could replace `IS_CORI=1` by
 * `echo "Please configure and build on cori* login nodes." && eixt 1`

If you think it's fine to configure and build on nid0*, I'll approve this.  But I have these questions for you.

gc00 (approved):
Approving in advance.  As we agreed, the code is fine, but let's echo a warning about slow builds on compute nodes.

JainTwinkle:
@gc00 
The `echo` command doesn't work without a rule in the `Makefile_config` file. I can use `info` instead of `echo` to print the warning message in this file. However, the message will be printed whenever the Makefile_config is included in other Makefiles. 
The `Makefile.in` in the root dir is a better place for such a warning.  So, I have updated the `Makefile.in` to give a warning if we are building MANA on a compute under the prereq rule.
Could you please review it again?

Issue #186, opened 2022-06-30
Problem with MPI_STATUS_IGNORE in MPI_Sendrecv


Marc-Miranda:
Good morning,

I was trying to run MANA with LAMMPS and encountered a problem with the implemented wrapper for MPI_Sendrecv. LAMMPS uses MPI_STATUS_IGNORE, but then in the wrapper it tries to dereference it, and the application does not crash but does not continue either. 
Changing
```
*status = sts[1];
```
for 
```
if( status != MPI_STATUS_IGNORE ){
     *status = sts[1];
}
```
solved the issue. I submit a minimal reproducible code reproducing the bug. Have a nice day!
[main.c.gz](https://github.com/mpickpt/mana/files/9019635/main.c.gz)

JainTwinkle:
Thanks, @Marc-Miranda. I have created a PR to include your suggested fix.

Marc-Miranda:
Hello! Sorry for bringing this up again. It seems to me there is an error in the Sendrecv wrapper. Shouldn't
```
if (status != MPI_STATUS_IGNORE || status != FORTRAN_MPI_STATUS_IGNORE) {
    *status = sts[1];
}
```
be 
```
if (status != MPI_STATUS_IGNORE && status != FORTRAN_MPI_STATUS_IGNORE) {
    *status = sts[1];
}
```
as is done at every other wrapper?

Best,
Marc

JainTwinkle:
@Marc-Miranda Thanks for catching. I will fix it today.

JainTwinkle:
@Marc-Miranda It's merged now!

PR #204, opened 2022-07-18
Do not set status if it's MPI_STATUS_IGNORE

JainTwinkle:
This is the fix suggested by Marc in issue #186.


gc00 (approved):
@JainTwinkle , I'm accepting this in advance, and it looks good to me.

Would you mind checking if there are other MPI functions where we forgot to check for MPI_STATUS_IGNORE?

Issue #202, opened 2022-07-16
Handling completed asynchronous request mappings


chirag-singh-memverge:
We are seeing a bug with CP2K where asynchronous request mappings are not being handled correctly. For example, we are observing the following sequence of events:

1. An Isend cal creates a new virtual --> real request mapping
2. This request is tested several times (normally)
3. The request completes, and we see Test_internal set the real request value to MPI_REQUEST_NULL (also expected behavior)
4. The next time the request is tested, the Test method removes the virtual-->real request mapping from our table
5. Wait is called on the virtual request, which no longer has a mapping. This calls the Test_internal to fail with an invalid request

How should this issue be approached? One potential issue is moving the checks for deleted from Test to Test_internal (see [here](https://github.com/mpickpt/mana/blob/main/mpi-proxy-split/mpi-wrappers/mpi_request_wrappers.cpp#L80)). However, this wouldn't solve the problem of multiple calls to test or wait on a completed request. In this case would checking for deleted mappings in Test_internal be the way to solve this issue?

chirag-singh-memverge:
@gc00 @xuyao0127 Would appreciate any input on this issue

PR #205, opened 2022-07-18
Fix logic error in request mapping for async requests

chirag-singh-memverge:


xuyao0127:
If a non-existing virtual request is passed into MPI_Test or MPI_Test_internal, there must be a bug in other parts of the program. I don't think we should try to convert it to MPI_REQUEST_NULL and hide the issue. Also, regarding your change in MPI_Test, if the request is MPI_REQUEST_NULL, VIRTUAL_ID_EXISTS will return false. As a result, codes that handle the null request will be skipped.

BTW, you need to add reviewers to your PR so that others can get notified to review your code.

chirag-singh-memverge:
If there is another issue that is causing this calling pattern, do you have any input on why we might be observing the behavior here: https://github.com/mpickpt/mana/issues/202

xuyao0127:
I think a better way is to set the virtual request to MPI_REQUEST_NULL on success. According to the MPI standard, the request parameter of MPI_Test and MPI_Wait is an in-and-out parameter. On success, the request should be set to MPI_REQUEST_NULL. So instead of making the virtual id translated to MPI_REQUEST_NULL, we can directly set the request parameter.

chirag-singh-memverge:
That seems reasonable, I'll give that a try.

mpi-proxy-split/mpi-wrappers/mpi_request_wrappers.cpp
```diff
@@ -47,11 +47,18 @@ int MPI_Test_internal(MPI_Request *request, int *flag, MPI_Status *status,
...
-    realRequest = VIRTUAL_TO_REAL_REQUEST(*request);
+      realRequest = VIRTUAL_TO_REAL_REQUEST(*request);
   }
   JUMP_TO_LOWER_HALF(lh_info.fsaddr);
   // MPI_Test can change the *request argument
   retval = NEXT_FUNC(Test)(&realRequest, flag, status);
+
+  // Set request to NULL on success to avoid issues with calling again
+  // with a bad virtual to real mapping
+  if(*flag){
```
gc00: `if(` -> `if (`

PR #214, opened 2022-07-19
Fix async request logic bug

chirag-singh-memverge:
See: https://github.com/mpickpt/mana/issues/202
PR #212 (initial attempt to fix issue, failed) - revert PR #205.
PR #205 (reverting bad request)


gc00 (approved):
LGTM.

PR #231, opened 2022-08-04
[MANA-196] Fixed issue when calling init thread by adding mana header

chirag-singh-memverge:
Adds a MANA Header to be saved in each checkpoint file. Currently, this header only saves the level of thread support used to initialize the thread. This header is saved to a file in each checkpoint directory, and is loaded on restart before initialization of the MPI process of each rank. 

This fix assumes that the same flag is being used to initialize all ranks. I couldn't find anything in the MPI standard to explicitly support this assumption see [here](https://www.mpi-forum.org/docs/mpi-4.0/mpi40-report.pdf) pg. 491. However, it seems like that in almost all cases this would be the case. We could try to verify this by saving each process thread support level in the DMTCP header also, and verifying on restart that the loaded level of thread support is equivalent to the saved value. This would allow us to abort if we restart with incorrect level of thread support.

### DIFF
--- a/mpi-proxy-split/lower-half/libproxy.c
+++ b/mpi-proxy-split/lower-half/libproxy.c
@@ -44,6 +44,7 @@
 #include "mpi_copybits.h"
 #include "procmapsutils.h"
 #include "lower_half_api.h"
+#include "../mana_header.h"
 
 LowerHalfInfo_t lh_info = {0};
 // This is the allocated buffer for lh_info.memRange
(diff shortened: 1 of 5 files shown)


chirag-singh-memverge:
BTW I am aware of the formatting issues, currently I'm looking mostly for feedback on the actual logic of this PR and I will fix the formatting issues when the PR logic is approved

gc00 (approved):
LGTM.

PR #236, opened 2022-08-09
[MANA-217] Convert freepass variable into semaphore, change commit_finish logic to match commit_begin

chirag-singh-memverge:
Converts freepass variable into a semaphore to avoid weird thread scheduling bug

Also changes commit_finish to run normally during presuspend phase

### DIFF
--- a/mpi-proxy-split/seq_num.cpp
+++ b/mpi-proxy-split/seq_num.cpp
@@ -18,7 +18,6 @@ extern int g_world_rank;
 extern int g_world_size;
 extern int p2p_deterministic_skip_save_request;
 volatile bool ckpt_pending;
-volatile bool freepass;
 volatile phase_t current_phase;
 unsigned int current_global_comm_id;
 reset_type_t reset_type;
@@ -26,6 +25,18 @@ reset_type_t reset_type;
 pthread_mutex_t seq_num_lock;
 sem_t user_thread_sem;
 sem_t ckpt_thread_sem;
+
+/* Freepass implemented using semaphore to avoid the following situation:
+*  T1: current_phase = STOP_BEFORE_CS;
+*  T2: freepass = true;
+*  while (current_phase == STOP_BEFORE_CS);
+*  T1: while (!freepass && ckpt_pending);
+*  freepass = false;
+*  current_phase = IN_CS;
+*  ....
+*  current_phase = STOP_BEFORE_CS;
+*/
+sem_t freepass_sem;
 
 std::map<unsigned int, unsigned long> seq_num;
 std::map<unsigned int, unsigned long> target_start_triv_barrier;
@@ -44,15 +55,17 @@ int comm_resume_target_reached(unsigned int global_comm) {
 
 void seq_num_init() {
   ckpt_pending = false;
-  freepass = false;
   pthread_mutex_init(&seq_num_lock, NULL);
   sem_init(&user_thread_sem, 0, 0);
   sem_init(&ckpt_thread_sem, 0, 0);
+  sem_init(&freepass_sem, 0, 0);
 }
 
 void seq_num_reset(reset_type_t type) {
   ckpt_pending = false;
-  freepass = false;
+
+  // Reset freepass_sem to 0, like in sem_init()
+  while (sem_trywait(&freepass_sem) == 0);
   reset_type = type;
   if (type == RESUME) {
     share_seq_nums(target_stop_triv_barrier);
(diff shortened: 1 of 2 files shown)


mpi-proxy-split/seq_num.cpp
```diff
@@ -170,16 +171,19 @@ void commit_begin(MPI_Comm comm) {
     // Stop before the critical section during checkpoint time.
     if (ckpt_pending && check_seq_nums()) {
       current_phase = STOP_BEFORE_CS;
-      while (!freepass && ckpt_pending);
-      freepass = false;
+      while(ckpt_pending){
```
gc00, line 174: QUESTION:  Isn't this the same as:
```
if (ckpt_pending) {
  sem_wait(&freepass_sem));
}
```
(And we could check the return value of sem_wait() if we're worried about that.)

mpi-proxy-split/seq_num.cpp
```diff
@@ -44,15 +44,15 @@ int comm_resume_target_reached(unsigned int global_comm) {
...
   pthread_mutex_init(&seq_num_lock, NULL);
   sem_init(&user_thread_sem, 0, 0);
   sem_init(&ckpt_thread_sem, 0, 0);
+  sem_init(&freepass_sem, 0, 0);
 }
 
 void seq_num_reset(reset_type_t type) {
   ckpt_pending = false;
-  freepass = false;
+  while(sem_trywait(&freepass_sem) == 0);
```
gc00, line 55: FORMATTING:  `while (sem_trywait ...)` with space after 'while'.

mpi-proxy-split/seq_num.cpp
chirag-singh-memverge: It's different, because of ckpt_pending becomes false after entering the loop then we will leave the loop, but if ckpt_pending becomes false after entering the if statement we would not exit then

mpi-proxy-split/seq_num.cpp
```diff
@@ -18,14 +18,14 @@ extern int g_world_rank;
...
 volatile bool ckpt_pending;
-volatile bool freepass;
 volatile phase_t current_phase;
 unsigned int current_global_comm_id;
 reset_type_t reset_type;
 
 pthread_mutex_t seq_num_lock;
 sem_t user_thread_sem;
 sem_t ckpt_thread_sem;
+sem_t freepass_sem;
```
gc00, line 28: Eventually, we'll want a comment above here.  If this is a semaphore, do we expect to give out two free passes, and then consume each individually?  Is that the motivation?  If so, do you have an example scenario?

mpi-proxy-split/seq_num.cpp
xuyao0127: @chirag-singh-memverge We can exit the loop at resume/restart time with a sem_trywait() and a sem_post().

mpi-proxy-split/seq_num.cpp
gc00: Can we move the comment to before the line:
```
sem_t freepass_sem;
```

mpi-proxy-split/seq_num.cpp
gc00: We still need a space after `while` in `while (sem_trywait ...)` on the line above.

mpi-proxy-split/seq_num.cpp
```diff
@@ -170,16 +182,15 @@ void commit_begin(MPI_Comm comm) {
     // Stop before the critical section during checkpoint time.
     if (ckpt_pending && check_seq_nums()) {
       current_phase = STOP_BEFORE_CS;
-      while (!freepass && ckpt_pending);
-      freepass = false;
+      while (ckpt_pending && sem_trywait(&freepass_sem) != 0);
```
gc00, line 185: Can we replace this by:
```
if (ckpt_pending) {
  sem_wait(&freepass_sem);
}
```
It's more like how we say it in English.  And when I see `sem_trywait()`, I assume the standard sem_post/sem_wait isn't sufficient, and we need an extension to the normal concept.  But here, the standard semaphore operations are enough.

mpi-proxy-split/seq_num.cpp
```diff
@@ -44,15 +55,15 @@ int comm_resume_target_reached(unsigned int global_comm) {
...
   ckpt_pending = false;
-  freepass = false;
   pthread_mutex_init(&seq_num_lock, NULL);
   sem_init(&user_thread_sem, 0, 0);
   sem_init(&ckpt_thread_sem, 0, 0);
+  sem_init(&freepass_sem, 0, 0);
 }
 
 void seq_num_reset(reset_type_t type) {
   ckpt_pending = false;
```
gc00, line 65: How about a comment:
```
// Re-initialize freepass_sem count to 0, as in original sem_init().
```

gc00 (approved):
Please make the minor suggested changes.  Otherwise, LGTM.

PR #241, opened 2022-08-17
Fix race condition in freepass

chirag-singh-memverge:
While the fix in #236 made it more rare, the race condition that was identified in that issue is still possible even with that fix. This change tries to fix that by enforcing that the thread in commit_begin must be notified by the thread giving the freepass that the thread giving the freepass has checked the state before changing its state to STOP_BEFORE_CS.


xuyao0127 (approved):
LGTM

PR #243, opened 2022-08-25
[MANA-213] Fixed previous typo in seqnum logic

chirag-singh-memverge:
This may be related to the issue we are seeing with the extremely slow checkpointing. 

This bug would cause the ranks to get stuck in the IN_CS state, and would never be set to IS_READY, which would make it impossible to checkpoint


JainTwinkle (approved):
Thanks, Chirag! LGTM!

PR #247, opened 2022-08-26
Fix sendrecv_replace wrappers

chirag-singh-memverge:
Resolved potential deadlock in Sendrecv_replace wrapper by forcing send/recv to occur in opposite order

### DIFF
--- a/mpi-proxy-split/mpi-wrappers/mpi_p2p_wrappers.cpp
+++ b/mpi-proxy-split/mpi-wrappers/mpi_p2p_wrappers.cpp
@@ -256,16 +256,37 @@ USER_DEFINED_WRAPPER(int, Sendrecv_replace, (void *) buf, (int) count,
                      (int) sendtag, (int) source,
                      (int) recvtag, (MPI_Comm) comm, (MPI_Status *) status)
 {
+  MPI_Request reqs[2];
+  MPI_Status sts[2];
 
-  // Send first
-  int retval;
-  retval = MPI_Send(buf, count, datatype, dest, sendtag, comm);
+  // Allocate temp buffer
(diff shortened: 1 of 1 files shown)


mpi-proxy-split/mpi-wrappers/mpi_p2p_wrappers.cpp
```diff
@@ -256,16 +256,35 @@ USER_DEFINED_WRAPPER(int, Sendrecv_replace, (void *) buf, (int) count,
...
 {
-
-  // Send first
   int retval;
-  retval = MPI_Send(buf, count, datatype, dest, sendtag, comm);
-  if (retval != MPI_SUCCESS) {
-    return retval;
+
+  // Send/recv must occur in different order to avoid deadlock
+  if (source > dest) { // Send first, then receive
```
JainTwinkle, line 262: Wouldn't the source and destination ranks for every rank participating in the MPI_Sendrecv_replace be same?

JainTwinkle: I meant, for rank 0 if send = 0 , dest = 1, then for rank 1 also it will be send = 0, dest = 1.

chirag-singh-memverge: checking

chirag-singh-memverge: Realized that source/dest aren't necessarily the same, fixed now to be independent of that (no deadlock either way)

JainTwinkle (approved):
Look good now!

PR #248, opened 2022-08-29
Don't reinit seq_num after comm free/recreate

xuyao0127:
This seems to be related to the long checkpointing problem.


xuyao0127:
MANA uses the global rank number of each member of a Group as the unique identifier of a communicator. So if a communicator is freed, but when another communicator with the same group is created, the new communicator will have the same identifier. In the wrapper of MPI_Comm_create, the sequence number of the newly created communicator will be initialized to 0. If there was a communicator that has the same group, its sequence number will be reset to 0. However, the target sequence number of the communicator is already set to a large number. All members of the communicator need to catch up to that target from 0, which will take a long time. We are suspecting this is the cause of the long checkpoint time problem instead of some stragglers that we thought before.

gc00 (approved):
LGTM.

PR #251, opened 2022-09-06
Change the mutex lock when ckpt INTENT arrives.

xuyao0127:
This may solve the occasional hanging issue during checkpoints. Previously, a process may share the sequence numbers and run into the critical section (lower half) before ckpt_pending is set to true in the ckpt thread. I move the mutex lock that protects sequence number sharing to protect both sequence number sharing and setting the value of ckpt_pending.


gc00 (approved):
LGTM.

PR #252, opened 2022-09-07
Fix Alltoall test

dahongli:


chirag-singh-memverge (approved):

gc00:
I have no objection to pushing this in.

But as a side comment, this doesn't change any of the semantics.  If one has a for loop nested inside another for loop, and they both use the index 'i', then any reference to 'i' in the inner scope must refer only to the inner index 'i'.

But maybe the renaming from 'i' to 'j' makes the code clearer for the reader.  :-)

PR #253, opened 2022-09-07
[MANA-235] Expanded file wrapper support

chirag-singh-memverge:
This solution takes a different approach from our earlier two strategies. Instead of relying on DMTCP to restore file descriptor characteristics and forcing the MPI_File to use the exact same file descriptor, we use the MPI API to exactly restore the file status instead. We can do this because we don't need the exact same file descriptors to be in use before and after restart for an MPI_File because the MPI_File struct is intended to be opaque, so there should be no difference in the user application between the two file descriptors, as long as the characteristics and content of the underlying file are the same. 

This is how we achieve this goal:
We maintain a global map of all open MPI_Files, g_params_map, for each process. Each time a file is created, it is virtualized, and added to this map, and when the file is closed, it is removed from the map. The relevant parameters for each file are stored in this map, and updated each time those parameters change. The one exception to this is the file offset, which is only set a single time immediately before checkpointing.

At checkpoint time, we write the content of this map to a file. 

At restart time, we restore the content of the map from the file. Then, for each entry in the map, we create a new MPI_File pointing to the same actual file. The characteristics of the map that were set are then restored from the parameter map. 

Major concerns: Translating constants to/from Fortran (MPI_Info constants, etc.)


mpi-proxy-split/mpi-wrappers/mpi_file_wrappers.cpp
```diff
@@ -42,19 +46,255 @@ USER_DEFINED_WRAPPER(int, File_open, (MPI_Comm) comm, (const char *) filename,
...
-DEFINE_FNC(int, File_get_size, (MPI_File) fh, (MPI_Offset *) size);
-DEFINE_FNC(int, File_read_at, (MPI_File) fh, (MPI_Offset) offset, (void *) buf,
-           (int) count, (MPI_Datatype) datatype, (MPI_Status *) status);
+USER_DEFINED_WRAPPER(int, File_get_size, (MPI_File) fh, (MPI_Offset *) size)
+{
+  int retval;
+  DMTCP_PLUGIN_DISABLE_CKPT();
+  MPI_File realFile = VIRTUAL_TO_REAL_FILE(fh);
+  JUMP_TO_LOWER_HALF(lh_info.fsaddr);
+  retval = NEXT_FUNC(File_get_size)(realFile, size);
```
JainTwinkle, line 128: Since you are saving the size every time `MPI_FIle_set_size` in `g_params_map[fh]._size`, do you need to do an actual call or can you just return the information from the global map?
In other words, are there other APIs that can change the size than `MPI_File_set_size` API?

chirag-singh-memverge: It appears other APIs can alter the size as well (such as writing to the file). I will change it to be treated the same as the file offset.

chirag-singh-memverge: Actually, this means it doesn't have to be saved and restored at all.

mpi-proxy-split/mpi-wrappers/mpi_file_wrappers.cpp
```diff
@@ -42,19 +46,255 @@ USER_DEFINED_WRAPPER(int, File_open, (MPI_Comm) comm, (const char *) filename,
...
+  MPI_File realFile = VIRTUAL_TO_REAL_FILE(fh);
+  MPI_Datatype realEtype = VIRTUAL_TO_REAL_TYPE(etype);
+  MPI_Datatype realFtype = VIRTUAL_TO_REAL_TYPE(filetype);
+  JUMP_TO_LOWER_HALF(lh_info.fsaddr);
+  retval = NEXT_FUNC(File_set_view)(realFile, disp, realEtype, realFtype,
+                                    datarep, info);
+  RETURN_TO_UPPER_HALF();
+  if (retval == MPI_SUCCESS) {
+    strncpy(g_params_map[fh]._datarep, datarep, strlen(datarep));
+    g_params_map[fh]._viewDisp = disp;
```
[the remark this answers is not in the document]
chirag-singh-memverge: I can do it at checkpoint

mpi-proxy-split/mpi-wrappers/mpi_file_wrappers.cpp
```diff
@@ -42,19 +46,295 @@ USER_DEFINED_WRAPPER(int, File_open, (MPI_Comm) comm, (const char *) filename,
...
+  MPI_File realFile = VIRTUAL_TO_REAL_FILE(fh);
+  MPI_Datatype realEtype = VIRTUAL_TO_REAL_TYPE(etype);
+  MPI_Datatype realFtype = VIRTUAL_TO_REAL_TYPE(filetype);
+  JUMP_TO_LOWER_HALF(lh_info.fsaddr);
+  retval = NEXT_FUNC(File_set_view)(realFile, disp, realEtype, realFtype,
+                                    datarep, info);
+  RETURN_TO_UPPER_HALF();
+
+  // Normally, we wait until checkpoint time to update any file characteristics
+  // However, we can't call set_view if there is no valid view
```
JainTwinkle, line 134: set_view or get_view?

chirag-singh-memverge: set_view
get_view completes, but gives garbage results

JainTwinkle: You'll be calling get_view during a checkpoint, and set_view during the restore, right?

Could you also update the comment clarifying that you will update the file characteristics in the global map, not in the MPI_File in the lower half?

JainTwinkle:
Your PR looks fine to me but I'll let @gc00 take a second look at your PR.

PS: please add a `FIXME` in `MPI_File_read/write_at_all` wrappers, stating that we assume this API is fast and doesn't need a trivial barrier in the hybrid-2PC form. However, if the application hangs here, one should include either a trivial barrier or an MPI_Barrier before moving forward.

chirag-singh-memverge:
@gc00 This PR is ready for review. It is passing all included tests.

gc00 (approved):
Let's push this in.  It's new functionality.  It will be easiest to test across different applications once it has been pushed in.

I like the design.  I think it's in the same spirit as the two earlier approaches.  The key, which you already note, is:
> The one exception to this is the file offset, which is only set a single time immediately before checkpointing.

Also, your design satisfies any concerns about ioctl().  If MPI_File_open does some magic call on ioctl() to set special attributes for the file descriptor, then your log-and-replay strategy will call MPI_File_open with the original parameters at the time of restart.  So, it all works.

PR #255, opened 2022-09-09
Fixed virtualization of certain file types

chirag-singh-memverge:
Minor bug fix


dahongli (approved):
Good catch. Thanks Chirag!

PR #256, opened 2022-09-09
Feature/type info api expansion

chirag-singh1:
Same as earlier PR, but moved from memverge github to my personal github

### DIFF
--- a/mpi-proxy-split/mpi-wrappers/generate-mpi-fortran-wrappers.py
+++ b/mpi-proxy-split/mpi-wrappers/generate-mpi-fortran-wrappers.py
@@ -24,7 +24,7 @@ else:
 declarations = declarations_file.read().split(';')[:-1]  # Each decl ends in ';'
 declarations_file.close()
 
-def print_mpi_type():
+def print_mpi_type_struct():
   print("#ifdef CRAY_MPICH_VERSION")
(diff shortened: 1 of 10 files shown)


gc00 (approved):
Since we're creating a new type here (e.g., `MPI_Type_create_hindexed'), don't we need to do log-and-replay on this?  Otherwise, on restart, the new MPI library won't know anything about the new type that was created prior to checkpoint.

If I'm right about this, then please add suitable comments with a `FIXME` explaining what is missing, and what still needs to be added.

But after you add the `FIXME`, it's fine to push this in.  I'm approving it in advance.

chirag-singh1:
We actually don't have to because the logging is handled by MPI_hindexed (see the implementation of create_hindexed)

gc00:
@chirag-singh1,
    Thanks for the comment.  Then this is ready to be pushed in (after rebasing).
Best,
- Gene

JainTwinkle:
@chirag-singh1 I believe the second commit is stopping git to rebase-and-merge this branch. Could you improve the PR to get rid of the merge commit and have a single clean commit?

chirag-singh1:
@JainTwinkle Should be fixed now

JainTwinkle:
@chirag-singh1 Thanks! The PR is already approved. You can go ahead and merge it.

PR #262, opened 2022-09-24
Remove redundant locks in MANA's virtual id table

xuyao0127:
DMTCP's virtual ID table is already protected by a lock. MANA's virtual ID table uses the DMTCP's virtual ID table and adds extra locks when operating the DMTCP table.


gc00 (approved):
LGTM.

PR #263, opened 2022-09-24
config --enable-fsgsbase-override; DMTP submodule

gc00:
MANA configure can do ./configure --enable-fsgsbase-override This passes the --enable-fsgsbase-override flag to the DMTCP submodule. This also updates the DMTCP submodule to include a commit for this new --enable-fsgsbase-override flag.

This is needed for MANA.  Otherwise, one configures and builds on the login computer, where the FSGSBASE patch for the kernel might be missing, and then tries to run on a computer node, which does have the FSGSBASE patch.

### DIFF
--- a/configure
+++ b/configure
@@ -616,8 +616,8 @@ enable_option_checking=no
 ac_subst_vars='LTLIBOBJS
 LIBOBJS
 PANDOC
-DEBUG
 subdirs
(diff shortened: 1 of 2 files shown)


karya0:
Closing it since we merged a dynamic change.

PR #267, opened 2022-10-05
Updated to new DMTCP KVDB api.

karya0:


mpi-proxy-split/mpi_plugin.cpp
```diff
@@ -361,16 +362,28 @@ computeUnionOfCkptImageAddresses()
...
-  dmtcp_global_barrier(kvdb.c_str());
+  dmtcp::kvdb::set(workerPath, "MANA_libsStart", libsStartStr);
+  dmtcp::kvdb::set(workerPath, "MANA_libsEnd", libsEndStr);
+  dmtcp::kvdb::set(workerPath, "MANA_highMemStart", highMemStartStr);
 
-  JASSERT(dmtcp_kvdb64_get(kvdb.c_str(), 0, (int64_t*) &minLibsStart) == 0);
-  JASSERT(dmtcp_kvdb64_get(kvdb.c_str(), 1, (int64_t*) &maxLibsEnd) == 0);
-  JASSERT(dmtcp_kvdb64_get(kvdb.c_str(), 2, (int64_t*) &minHighMemStart) == 0);
+  string kvdb = "/plugin/MANA/CKPT_UNION";
+  dmtcp::kvdb::request64(dmtcp::kvdb::KVDBRequest::MIN, kvdb, "libsStart", (int64_t) libsStart);
```
JainTwinkle, line 375: @karya0 could you remind me if we need to check the return value from request64 or not?

mpi-proxy-split/seq_num.cpp
```diff
@@ -274,14 +277,21 @@ void drain_mpi_collective() {
...
-    dmtcp::string csId = "MANA-CRITICAL-SECTION-" + jalib::XToString(round_num);
-    dmtcp::string convergeId = "MANA-CONVERGE-" + jalib::XToString(round_num);
-    dmtcp_kvdb64(DMTCP_KVDB_INCRBY, convergeId.c_str(), 0, check_seq_nums(false));
-    dmtcp_kvdb64(DMTCP_KVDB_OR, csId.c_str(), 0, current_phase == IN_CS);
-    dmtcp_global_barrier(barrierId.c_str());
-    dmtcp_kvdb64_get(convergeId.c_str(), 0, &num_converged);
-    dmtcp_kvdb64_get(csId.c_str(), 0, &in_cs);
+    char key[32] = {0};
+    char barrier_id[32] = {0};
+    dmtcp::string cs_id = "/plugin/MANA/CRITICAL-SECTION";
```
JainTwinkle, line 282: If we're not changing these ID strings ever, let's make them `constexpr char *` like `comm_seq_max_db`.

mpi-proxy-split/seq_num.cpp
```diff
@@ -274,14 +277,21 @@ void drain_mpi_collective() {
...
+    dmtcp::string converge_id = "/plugin/MANA/CONVERGE";
+    sprintf(barrier_id, "MANA-PRESUSPEND-%06d", round_num);
```
JainTwinkle, line 284: I'm not familiar with the seq algorithm, should it be `MANA/PRESUSPEND-%06d` like `plugin/MANA/CRITICAL-SECTION`?

mpi-proxy-split/mpi_plugin.cpp
karya0: Good catch. Added an assert.

mpi-proxy-split/seq_num.cpp
karya0: Done.

JainTwinkle (approved):
LGTM, please test the PR before merging in.
Thanks!

PR #268, opened 2022-10-06
[MANA-61] Switch from upper half when processing signals.

karya0:


karya0:
@JainTwinkle : Added the #ifdef in the code now. Please take a look.

mpi-proxy-split/mpi_plugin.cpp
```diff
@@ -149,6 +154,93 @@ dmtcp_skip_memory_region_ckpting(const ProcMapsArea *area)
...
+  if (signum == dmtcp_get_ckpt_signal() ||
+      signum >= MaxSignals) {
+    return NEXT_FNC(sigaction)(signum, act, oldact);
+  }
+
+  if (act != NULL) {
+    struct sigaction newAct = *act;
+    if (newAct.sa_flags & SA_SIGINFO) {
+      newAct.sa_sigaction = mana_signal_sa_sigaction_wrapper;
+      ret = NEXT_FNC(sigaction)(signum, &newAct, oldact);
```
JainTwinkle, line 216: oldact -> oldAct for consistency.

mpi-proxy-split/mpi_plugin.cpp
```diff
@@ -461,17 +553,40 @@ mpi_plugin_event_hook(DmtcpEvent_t event, DmtcpEventData_t *data)
   switch (event) {
     case DMTCP_EVENT_INIT: {
       JTRACE("*** DMTCP_EVENT_INIT");
+      JASSERT(dmtcp_get_real_tid != NULL);
+      initialize_signal_handlers();
       initialize_segv_handler();
       JASSERT(!splitProcess()).Text("Failed to create, initialize lower haf");
```
JainTwinkle, line 559: Let's fix a typo here!
haf -> half

mpi-proxy-split/mpi_plugin.cpp
```diff
@@ -461,17 +553,40 @@ mpi_plugin_event_hook(DmtcpEvent_t event, DmtcpEventData_t *data)
...
     }
     case DMTCP_EVENT_EXIT: {
       JTRACE("*** DMTCP_EVENT_EXIT");
       seq_num_destroy();
       break;
     }
+
+    case DMTCP_EVENT_PTHREAD_START: {
+      // Do we need a mutex here?
+      DmtcpMutexLock(&g_upper_half_fsbase_lock);
```
JainTwinkle, line 575: tid would be unique for each thread. So, I guess we can skip the mutex here.

mpi-proxy-split/mpi_plugin.cpp
```diff
@@ -461,17 +553,40 @@ mpi_plugin_event_hook(DmtcpEvent_t event, DmtcpEventData_t *data)
...
+      g_upper_half_fsbase[dmtcp_get_real_tid()] = getFS();
+      DmtcpMutexUnlock(&g_upper_half_fsbase_lock);
+      break;
+    }
+
+    case DMTCP_EVENT_PTHREAD_RETURN:
+    case DMTCP_EVENT_PTHREAD_EXIT: {
+      // Do we need a mutex here?
```
JainTwinkle, line 583: Same as above!

mpi-proxy-split/mpi_plugin.cpp
karya0: STL `unordered_map` is not thread-safe so that's why we need a mutex.

JainTwinkle: I believe different threads can access and modify different elements concurrently. In our case, unique TIDs should be enough to ensure that no two threads will access the same key.

karya0: They can access it but if one thread tries to modify unordered_map (insert, erase), that'll cause a race condition and the results are unexpected. In multi-threaded application, STL containers need to be protected with mutex.

JainTwinkle: Got it.

yuelimv (approved):

PR #269, opened 2022-10-08
Disable fsbase switch in signal handlers.

karya0:


karya0:
Closing since #270 handles it better.

PR #270, opened 2022-10-09
Added dynamic check for fsgsbase.

karya0:


mpi-proxy-split/split_process.h
```diff
@@ -36,35 +45,36 @@
 static inline unsigned long getFS(void)
 {
   unsigned long fsbase;
-#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,9,0) || defined(HAS_FSGSBASE) || defined(ENABLE_FSGSBASE_OVERRIDE)
```
JainTwinkle, line 39: If we're switching to dynamic then do we still need HAS_FSGSBASE and ENABLE_FSGSBASE_OVERRIDE config options?

karya0: We don't need those options. However, I haven't removed them in this PR due to regression scripts. Will create a separate PR once the test scripts are updated (after this PR is merged).

JainTwinkle (approved):
LGTM! :)

mpi-proxy-split/split_process.h
gc00: @karya0 and @JainTwinkle ,
    Can we restore the part of the line that says Linux 5.9 and later have FSGSBASE?
    And I'd stlil like to keep the ENABLE_FSGSBASE_OVERRIDE for testing of MANA.
The line that I want to restore is here in `mpi-proxy-split/split_process.h`:
```
#if LINUX_VERSION_CODE >= KERNEL_VERSION(5,9,0) || defined(HAS_FSGSBASE) || defined(ENABLE_FSGSBASE_OVERRIDE)
```
The rationale is that the way Linux represents its HWCAP may or may not change in the future.  I can imagine them deciding to have two variants based on strong or weak support for this feature.  But Linux is unlikely to remove the support for FSGSBASE after Linux 5.9.
    I think it's a mistake to do this kind of "defensive programming", in which Linux adds a feature at some version, and maintains it for backwards compatibility, but we'll test anyway to see if Linux decided to abandon backwards compatibility.  If FSGSBASE stops working after Linux 5.9, it's better if MANA breaks, so we can look carefully at what happened i a later Linux --- instead of doing the defensive programming that may or may not cover up a lack of backwards compatibility.

gc00 (commented):
Please see my comment about keeping the idea that Linux 5.9 and later support FSGSBASE.  I think it's a mistake to assume some future-proof standard for HWCAP that will dynamically test this feature forever.

If we start to do that for FSGSBASE, there are lots of places where we can assume that Linux will abandon backwards capability and we need to dynamically test forevver to see if Linux has abandoned backwards compatibility.

PR #274, opened 2022-10-13
Added signal() wrapper.

karya0:


JainTwinkle (approved):
LGTM!
Should we move signal handler-related code in a separate file?

karya0:
Yeah, at some point, we should sanitize mpi_plugin.cpp.

PR #275, opened 2022-10-14
Added missing disable/enable ckpt in WaitAny.

karya0:


xuyao0127 (approved):
LGTM

karya0:
@JainTwinkle : I jumped the gun a few seconds too soon. Did you mean something like: https://github.com/mpickpt/mana/pull/276?

JainTwinkle:
@karya0 yes! #276 looks correct to me. @xuyao0127 do you agree?

xuyao0127:
I think @JainTwinkle is correct. We need to re-enable checkpoint after counters are updated. And @karya0 your new PR still has some problem. The second EANBLE_CKPT is after a return statement. So if it's returned, then it cannot re-enable checkpoint.

PR #276, opened 2022-10-14
Improvement to enable/disable ckpt in WaitAny.

karya0:


mpi-proxy-split/mpi-wrappers/mpi_request_wrappers.cpp
```diff
@@ -289,6 +289,7 @@ USER_DEFINED_WRAPPER(int, Waitany, (int) count,
         *local_index = i;
         return retval;
       }
+      DMTCP_PLUGIN_ENABLE_CKPT();
```
JainTwinkle, line 292: There are multiple returns. I'm afraid that the function can return before enabling CKPT.

xuyao0127: Yes, I have the same concern here.

JainTwinkle: We can use break instead of return.

karya0: Good catch. Fixed.

JainTwinkle (approved):

PR #278, opened 2022-10-18
Added MANA_FILE_REGEX env var.

karya0:
MANA plugin will use this regex to open files in O_APPEND|O_WRONLY mode at ckpt so that they can be properly truncated on restart.


xuyao0127 (approved):
LGTM

PR #281, opened 2022-11-03
MTCP mmap_fixed_noreplace updates.

karya0:


mpi-proxy-split/mpi_plugin.cpp
```diff
@@ -198,59 +292,122 @@ closeCkptFileFds()
...
+    return true;
+  }
+
+  return false;
+}
+
+static bool
+isLhRegion(const ProcMapsArea *area)
+{
+  if (isBlacklistedRegion(area) ||
```
gc00, line 378: Suggest: `Blacklisted` -> `Whitelisted` (as above)

mpi-proxy-split/mpi_plugin.cpp
```diff
@@ -88,34 +88,125 @@ mana_state_t mana_state = UNKNOWN_STATE;
...
+static bool isLhMmapRegion(const ProcMapsArea *area);
+static bool isLhMpiInitRegion(const ProcMapsArea *area);
+static bool isLhRegion(const ProcMapsArea *area);
+
+static inline int
+regionContains(const void *haystackStart,
+               const void *haystackEnd,
+               const void *needleStart,
+               const void *needleEnd)
+{
```
gc00, line 110: Suggest adding comment:
```
  // Does haystack region contain needle region?
```

gc00 (approved):
I have some minor, local changes suggested.
Otherwise, I'm approving this in advance.

gc00:
@karya0 ,
    I've approved your PR #281, with some minor (trivial) changes requested before pushing it in.  Could you make those changes and push it in this afternoon, before Tom's QA script starts running?  Thanks.

PR #282, opened 2022-11-08
[MANA-277] Ignore /anon_hugepages when computing min libs addr.

karya0:
Otherwise, we miscalculate libsStart and restart fails to reserve the address range.

### DIFF
--- a/mpi-proxy-split/mpi_plugin.cpp
+++ b/mpi-proxy-split/mpi_plugin.cpp
@@ -644,7 +644,14 @@ computeUnionOfCkptImageAddresses()
       highMemStart = area.addr;
     }
 
-    if (area.addr > heapAddr && !isLhRegion(&area)) {
+    if (area.addr > heapAddr &&
+        !isLhRegion(&area) &&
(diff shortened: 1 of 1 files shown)


dahongli (commented):
Looks good. Thanks Kapil!

dahongli (approved):

gc00 (approved):
This looks like a great fix. I'm approving it in advance.
But please add a comment about what an "/anon_hugepage" segment represents.
Is "/anon_hugepage" associated with the lower half? Or is it a memory region that we will somehow want to ignore later? Who created "/anon_hugepage"? (Was it created by GNI, or libmpich.a, or other?

Whatever comment you can add here will be very helpful to us a year from now, when we've forgotten the details of the code.

_And in general, thanks very much for this cleanup in analyzing the different memory regions!
All of this is tremendously important._

We needed to know things like what is lower-half versus upper-half, and what might be created by the GNI network, possibly even at runtime. And we didn't have enough time to analyze this carefully, because we were fighting multiple battles.

PR #284, opened 2022-11-11
[MANA-277] Fixed libs-start calculation.

karya0:
In some rare cases, `strstr(path, ".so")` would match with a temp file containing ".so" in the name -- thus throwing off our libs-start calculations. This PR makes the heuristics a bit stronger.


JainTwinkle (approved):
LGTM!

PR #285, opened 2022-11-17
Added MANA_SEGV_DEBUG_LOOP env to enable segfault handler.

karya0:
Do not install segfault handler by default. One can set MANA_SEGV_DEBUG_LOOP to 1 to install the handler.


xuyao0127 (approved):
LGTM

PR #286, opened 2022-11-18
Do not update index if WaitAny call fails.

karya0:


mpi-proxy-split/mpi-wrappers/generate-mpi-fortran-wrappers.py
```diff
@@ -158,7 +158,9 @@ def emit_wrapper(decl, ret_type, fnc, args, arg_vars):
           warnings.warn("'int* index' found in {fnc} - likely to cause erroneous behavior".format(fnc=fnc))
         # We need this line because Fortran arrays start from 1, and *index
         # refers to the index of the first request that completes in an array.
-        print("  *local_index = *local_index + 1;")
+        print("  if (*ierr != MPI_UNDEFINED) {")
```
dahongli, line 161: Should we check if `*local_index != MPI_UNDEFINED`? Or we can further tight the condition as both *ierr is MPI_SUCCESS and check *local_index is not undefined?

dahongli: In my experiment, the index is UNDEFINED and *ierr is zero. So in that case at least, we don't want to increase index. 
I am not sure what the caller behavior is if ierr is not zero. I am wondering if the non-zero error would trigger error handler.

xuyao0127 (approved):
LGTM

PR #291, opened 2022-12-13
[MANA-285] Erase MPI_Type_* record from record-replay

dahongli:
For each MPI_Type_* operation, a record is added to record-replay record list. Even if the MPI_Type is freed, the record is still kept. That consumes lots of memory and eventually cause out-of-memory for some cp2k workload.

This change erases the MPI_Type record when it is freed and there is no new type depends on it. The following example and workflow illustrates the solution. The example shows new type depends an old type we created.
1] MPI_Type_hvector creates a new type A from the vector of pre-defined MPI_INT.
2] MPI_Type_hvector creates a new type B from the vector of the type A which is created in step #1 
3] MPI_Type_free frees type A. Please note we can't immediately free type A record since the type B still depends on it when we replay the record.
4] MPI_Type_free frees type B.

We could not immediately free "type A" record in step #3 since the "type B" record still depends on the type A record. To solve this issue, we add a reference count for each "type" record. When a new type record is created, we set the reference count to "1". The reference count of the old type is increased when a new type is created from the old type. When a type is freed, its reference count is decreased. If its reference drops to zero, the type is freed. In addition, we decreases the reference count of each its old type and free the old type if the ref count of old type drops to zero. This procedure goes recursively until all old types are pre-defined type.

Future work: This change only frees the record of three MPI type operations, MPI_Type_hvector, MPI_Type_commit and MPI_Type_free. They are used by cp2k auslab workload test. In my next change, all remaining MPI_Type operations will be supported.

### DIFF
--- a/mpi-proxy-split/record-replay.h
+++ b/mpi-proxy-split/record-replay.h
@@ -383,16 +383,17 @@ namespace dmtcp_mpi
         MpiRecord *rec = new MpiRecord(cb, type, (void*)fPtr);
         if (rec) {
           rec->addArgs(args...);
-          MPI_Request req = MPI_REQUEST_NULL;
 	  // All collective calls are translated in MANA to the async calls Ibarrier/Ireduce/Ibcast.
 	  // If MANA uses other async calls, we need other cases.
 	  switch (type) {
             case GENERATE_ENUM(Ibarrier):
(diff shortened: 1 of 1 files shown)


xuyao0127 (approved):
LGTM

PR #292, opened 2022-12-13
Remove MPICH_SINGLE_COPY_OFF to enable xpmem

xuyao0127:
MPICH_SINGLE_COPY_OFF was set to disable xpmem. Older versions of MANA can not tag xpmem regions in the memory correctly so we need to disable it to correctly checkpoint/restart. But disabling xpmem will slow down communications of large data sets. Now MANA is able to handle xpmem so we can remove this environment variable.


dahongli (approved):
The change looks good. Thanks Yao!

Just curious, any idea which commit adds support of xpmem to MANA?

xuyao0127:
@dahongli We don't know which commit specifically adds the support. We think it's about the overall improvement in tagging memory regions in the upper and lower halves.

PR #298, opened 2023-02-10
Add assertions in P2P draining, abort if timeout

xuyao0127:
Abort if draining P2P timeout or there's pending MPI_Isend at the point of a checkpoint. If the draining process fails, there will be messages lost on restart. In this case, we prefer to terminate the program. Otherwise, we will have incorrect results after restarting.


gc00 (approved):
LGTM.

PR #304, opened 2023-03-16
Use GSS to compile gethostbyname

dahongli:


jiamingz9925 (approved):
looks good

mpi-proxy-split/lower-half/gethostbyname-static/Makefile
```diff
@@ -12,10 +13,10 @@ gdb: a.out
 	gdb --args ./a.out localhost
 
 a.out: gethostbyname_static.c gethostbyname_proxy
-	${CC} ${CFLAGS} -DSTANDALONE $< -o $@
+	${GCC} ${CFLAGS} -DSTANDALONE $< -o $@
```
gc00, line 16: With the above chance, we can still use the more conventional `${CC}`.

mpi-proxy-split/lower-half/gethostbyname-static/Makefile
```diff
@@ -12,10 +13,10 @@ gdb: a.out
...
 
 gethostbyname_proxy: gethostbyname_proxy.c
-	${CC} ${CFLAGS} $< -o $@
+	${GCC} ${CFLAGS} $< -o $@
```
gc00, line 19: With the above chance, we can still use the more conventional `${CC}`.

mpi-proxy-split/lower-half/gethostbyname-static/Makefile
```diff
@@ -1,3 +1,4 @@
+GCC=gcc
```
gc00, line 1: I suggest simply:  `CC=gcc` and continuing to use `CC` everywhere below.  This makes the change local.

Let's also add a small comment above it:
```
# This is needed for a statically linked MANA lower-half, in order to permit the statically linked
# code to use 'gethostbyname()'  We require 'gcc' here for this hack.
# When the MANA lower half is dynamically linked, all of this will go away.
```

gc00 (approved):
@dahongli, I'm approving this in advance.  Please make the suggested changes.

PR #306, opened 2023-04-07
fix static lib link issue on perlmutter

jiamingz9925:
This pull request is to resolve the build failure error for MANA on Perlmutter.
Now MANA can be built successfully on Perlmutter and Cori from my end.
Please verify this fix can resolve the issue on other's node as well.


dahongli (approved):
Look good. Thanks Jiaming!

PR #309, opened 2023-04-27
Jiaming/mpi logger

jiamingz9925:
A mpi trace tool that allows people to log mpi calls without running mana. The user program can run natively and generate a mpi trace log to stdout.
Which mpi calls to trace needs a manual adjustment.
To use it just build with `make`


mpi-proxy-split/dbg_logger/Makefile
```diff
@@ -0,0 +1,11 @@
+default: fortran_constants.o get_fortran_constants.o
+	CC -I../../include -fPIC -c -g3 -o mpi_logger.o mpi_logger.cpp -lmpi -O0
+	CC -shared -fPIC -o liblogger.so mpi_logger.o get_fortran_constants.o fortran_constants.o
+clean:
+	rm -f -v *.o *.so
+
+fortran_constants.o: fortran_constants.f90
+	ftn -g3 -O0 -fPIC -c -o $@ $<
```
gc00, line 8: Change the line, above, to use:
  `${MPIFORTRAN}` instead of `ftn`

gc00 (approved):
I've added some minor changes for you to insert, before committing this.

Don't you also need the wrappers for the various MPI functions, or is that coming in a later PR?

jiamingz9925:
Are you suggesting mpi wrappers being used in mana run? I thought this pr should only focus on the logger tool for native run. @gc00

gc00:
@jiamingz9925 , I agree that this is only for native MPI, and not MANA.  But when you build liblogger.so, doesn't it include wrapper functions for MPI calls that then pass the call directly to the MPI library, and then print the arguments/results in the wrapper, before the wrapper returns.  MPI_Reduce was one of the wrapper examples that we looked at.

I'm not weeing any wrapper for `MPI_Reduce` in this logger for native MPI.

jiamingz9925:
Its in mpi_logger.cpp, I believe it is included, you might need to load diff manually since its too long. @gc00

mpi-proxy-split/dbg_logger/mpi_logger_utils.h
```diff
@@ -0,0 +1,142 @@
+#ifndef _MPI_LOGGER_UTILS_H
+#define _MPI_LOGGER_UTILS_H
+
+#include <mpi.h>
+
+#define EXTERNC extern "C"
+#define ENABLE_LOGGER_PRINT true
```
gc00, line 7: Please change `true` to `1`, above.  That's more conventional for the C preprocessor.

Actually, `#if ENABLE_LOGGER_PRINT` will simply check if this is t to `0` or any non-zero value.  Probably, even `false` counts as a true value.  The preprocessor works with any language -- not just C++.

An alternate from is `#ifdef ENABLE_LOGGER_PRINT`.  In that form, it simply checks if you did even `#define ENABLE_LOGGER_PRINT` (with no value for the preprocessor variable.  In that case, you comment it out if you don't want it, and even `0` counts as a _defined_ value.

PR #316, opened 2023-05-11
Reserve stack memory region at checkpoint restart

dahongli:
Stack region is usually starting at minmum high memory region. Stack is less than 8MB. Reserve this 8MB region. Otherwise, GNI driver may use it when we need to map stack region to this memory address. Please note, so far MTCP restart only takes minHighMemStart. It does not have information about the high memory region end address

### DIFF
--- a/mpi-proxy-split/mpi_plugin.cpp
+++ b/mpi-proxy-split/mpi_plugin.cpp
@@ -97,7 +97,8 @@ void *heapAddr = nullptr;
 // #define DEBUG
 
 #undef dmtcp_skip_memory_region_ckpting
-const VA HIGH_ADDR_START = (VA)0x7fff00000000;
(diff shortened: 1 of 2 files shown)


gc00 (approved):
Thanks for this fix to adapt to Perlmutter!

I'm approving it now.  I have one question:
  Does it work if you modify _only_ `mpi-proxy-split/mpi_plugin.cpp` to set `VA HIGH_ADDR_START` to `0x7ffc00000000`?  Could you test with just that one change?

If that is enough to make it work, then we should not also change `mtcp_restart_plugin.c` to use `start2`/`end2`.

Could you also do that test, and add a comment on this PR?

Meanwhile, I'm approving so that you can push in whichever version of the PR is better (the smaller change is preferred if that works).  Thanks.

PR #317, opened 2023-05-22
[WIP] Correctness - Feature - Debugging Library (libreproducible.so) for Enabling Reproducibility in Native Runs.

tarunsmalviya:


mpi-proxy-split/dbg_reproducible/mpi_reproducible.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+
+/* We use 'static' becuase we don't want the overhead of the compiler
+initializing these to zero each time the function is called. */
+#define MAX_ALLREDUCE_SENDBUF_SIZE (1024 * 1024 * 16) /* 15 MB */
+
+int
+MPI_Init_thread(int *argc, char ***argv, int required, int *provided)
+{
+
```
jiamingz9925, line 20: In order to use Fortran constants, you need to call something like `get_fortran_constants()` on initialization method like this one.

gc00:
@tarunsmalviya , Can you finish this PR?  Thanks.

PR #318, opened 2023-05-31
Fixed double-free memory corruption.

karya0:


mpi-proxy-split/record-replay.h
```diff
@@ -106,34 +107,22 @@ namespace dmtcp_mpi
...
-      // by CREATE_LOG_BUF in MPI_Cart functions.
-      _type = dmtcp_mpi::TYPE_INT_ARRAY;
       if (_data && data) {
-        memcpy(_data, data, len);
-      }
-    }
-
-    ~FncArg()
-    {
-      if (!_data) {
```
gc00, line 135: This doesn't matter now that we are revising this code.  But isn't `if (!_data)` a separate original bug (a memory leak)?

First of all, `_data` will always be a valid pointer.  All versions of the constructor allocated a _data buffer.  Maybe this logic is left over from an earlier design.  So, it seems like `!_data` will always be false, and so we never free `_data`.  This is a memory leak.

mpi-proxy-split/record-replay.h
```diff
@@ -106,34 +107,22 @@ namespace dmtcp_mpi
...
+    // by CREATE_LOG_BUF in MPI_Cart functions.
+    FncArg(const void *data, size_t len, dmtcp_mpi::TYPE type = TYPE_INT_ARRAY)
+      : _type(type)
     {
-      _type = type;
-      if (_data && data) {
-        memcpy(_data, data, len);
+      void *ptr = JALLOC_HELPER_MALLOC(len);
+      if (ptr) {
+        _data = std::shared_ptr<void>(ptr,
```
gc00, line 120: I suggest an added comment here:
````
          // FIXME:  This 'shared_ptr' fixes an apparent double-free.
           //   However, this is a temporary fix, until we can discover
           //     why this happens.
```

gc00 (approved):
I'm approving now, in order to get this into MANA quickly.  But if I understand the issue properly, then I have a question.

@karya0, You report that you're seeing a double free.  In the code in this file alone, there is no obvious double-free.  Could you say more about where you are seeing the double-free?

Could it be that we are explicitly calling the FncArg destructor somewhere, and then implicitly calling it again when we destroy a higher level object that uses FncArg?  But I don't see any explicitly destructor of FncArg.  So, maybe this happens inside a higher-level object.  If you can point to where the double-free is, that will be helpful.

@karya0, @jiamingz9925, @dahongli :  Can one of you insert some logic inside `~FncArg` (the destructor) to see if it is being called twice?  An obvious way of testing is to add to the constructor a true field: `: inUse=true`.  And then in `~FncArg`, check if `inUse` is already false (an error), or sel set inUse to be `false'.  Thanks.

I can guess how this bad design might have happened.  Originally, a FncArg was created only during regular launch/resume, and it was only during restart that we would destroy a FncArg.  This probably did not have any malloc bugs.  If so, this bug would be exhibited only in a complex use of MPI_Isend/Irecv, etc.  Our simple tests would not show this.

Later, @xuyao0127 and I modified the logic of `FncArg` to create it temporarily during launch, and then destroy it again later during launch, once we could verify that we would never have to replay it.  We were doing this for MPI_Send/Recv, or maybe MPI_Isend/Irecv.  But I'm not sure about the details.  It was a year ago.

@xuyao0127, can you also review our use of `FncArg`?  And I'll try to do that too.

gc00:
@karya0 , Is this PR #318 still valid, or should it be closed without committing?  Thanks.

PR #320, opened 2023-06-09
Check if a memory region

dahongli:
This change checks if a memory region is overlapped with the existing region. If they are overlapped, then we do not need to reverve it.

### DIFF
--- a/restart_plugin/mtcp_restart_plugin.c
+++ b/restart_plugin/mtcp_restart_plugin.c
@@ -1,3 +1,4 @@
+#include <stdbool.h>
 #include <asm/prctl.h>
 #define _GNU_SOURCE // needed for MREMAP_MAYMOVE
 #include <sys/mman.h>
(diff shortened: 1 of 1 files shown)


dahongli (commented):
test, please

dahongli (commented):
test this please

gc00 (approved):
I think the test for overlap is not complete enough.  Here is a function that we can include that tests for all types of overlap.  For example, right now, we could have:
  start1 < start2 < end1 < end2
and your code would not diagnose this as an overlap.

Here is an overlap functio produced by Google Bard that we can use.
Please insert the `is_overlap` function into the code, and call that functio to test for overlap.
```
int is_overlap(char *start1, char *end1, char *start2, char *end2) {
  // Check if either memory region is NULL (set to 0).
  if (start1 == NULL || start2 == NULL) {
    return false;
  }
  // Check if the two memory regions are contiguous
  if (end1 < start2 || end2 < start1) {
    return false;
  }
  // The two memory regions overlap if the end of one region is
  //   greater than or equal to the start of the other region
  return end1 >= start2 || end2 >= start1;
}
```

I'm approving the PR in advance.  Please use this `is_overlap()` function in the PR.

jackpao-memverge:
test this please

PR #323, opened 2023-06-21
Robust lower-half mmap()

gc00:
**[This PR has now been updated.  It no longer mmap's the lhMemRange in advance.  @jiamingz9925 reports that this allows the PR to succeed with VASP6/Si256, when executing with  `MPICH_SMP_SINGLE_COPY_SIZE=819200`.  We should test this again, after PR #328 is pushed in.]**

If this PR succeeds, we can push it int, and then go back and analyze why mmap'ing the lhMemRange causes a problem.  Reserving lhMemRange in advance may be necessary in order to avoid using `MPICH_SMP_SINGLE_COPY_SIZE=819200`.  That would allow for greater use of `/dev/xpmem` segments, which can improve performance.

And this raises 2 GB to 8 GB in a commit, to increase the gap between lower and upper half.

I'm attaching the text from the most important commit in this PR.

 Make lower-half mmap/munmap more robust:
1. libxpmem (and maybe other kernel drivers) can use ioctl to mmap
        additional memory segments, such as /dev/xpmem.  If they mmap
        many such segments or a few large ones, they can occupy a lot
        of the lower-half lh_memRange (aka lh_mem_range in split_process.cpp).
2. This can cause a problem either:  because the lh_memRange does not
        have enough empty space for the lower half to mmap; or more often
        when there are many upper-half libraries, whose mmap allocations
        grow downward and enter the lh_memRange.
3. The lower half adds MAP_FIXED to its mmap's, thinking that it controls
        all of lh_memRange, and trying to allocate its mmaps contiguously
        and growing upward in lh_memRange.  Its mmap can receive an ENOMEM
        if it tries to clobber /dev/xpmem.  It then abandons the mmap on
        the assumption that the caller to mmap (libmpi) would want to see
        the ENOMEM.  The libmpi caller can then test another region with
        mmap (maybe with mmap and a specified address or with mmap/MAP_FIXED
        and a specified address, followed by munmap).  The lower-half mmap
        adds the MAP_FIXED flag, as usual.  This can cause the lower-half
        mmap to clobber an upper-half mmap, if the upper-half mmap had
        been allocated in the lh_memRange.  (The upper-half mmap may enter
        the lh_memRange if the upper half has very many libraries, since
        libraries are populated using mmap without a suggested address,
        and those mmap's grow downward.)
4. The (partial) solution is for the lower-half mmap to use
        MAP_FIXED_NOREPLACE instead of MAP_FIXED, and test for EEXIST.
        Additionally, it may be necessary in a future commit to choose a
        different lh_memRange further away from the many upper-half libraries
        that are being mmap'ed.

### DIFF
--- a/mpi-proxy-split/lower-half/mmap64.c
+++ b/mpi-proxy-split/lower-half/mmap64.c
@@ -17,6 +17,7 @@
    License along with the GNU C Library; if not, see
    <http://www.gnu.org/licenses/>.  */
 
+#include <linux/version.h>
 #include <errno.h>
 #include <string.h>
 #include <unistd.h>
@@ -28,6 +29,8 @@
 #include "mmap_internal.h"
 #include "mpi_copybits.h"
 #include <sys/syscall.h>
+
+#define HAS_MAP_FIXED_NOREPLACE LINUX_VERSION_CODE >= KERNEL_VERSION(4, 17, 0)
 
 #ifndef __set_errno
 # define __set_errno(Val) errno = (Val)
@@ -59,13 +62,15 @@
 // Modified to keep track of regions mmaped by lower half
 
 // TODO:
-//  1. Make the size of list dynamic
+//  1. Add byte, 0b10101010, everywhere to lh_memRange, when first
+//     calling mmap within 'if (!initialized)' logic, below.
+//  2. Make the size of list dynamic
 
 // Number of valid objects in the 'mmaps' list below
 int numRegions = 0;
 
 // List of regions mmapped by the lower half
-MmapInfo_t mmaps[MAX_TRACK] = {0};
+MmapInfo_t mmaps[MAX_TRACK] = {{0}};
 
 // Pointer to the next free address used for allocation
 void *nextFreeAddr = NULL;
@@ -118,9 +123,7 @@ getNextAddr(size_t len)
   nextFreeAddr = (char*)curr + ROUND_UP(len) + PAGE_SIZE;
 
   // Assert if we have gone past the end of the lower half memory range
-  if (nextFreeAddr > lh_memRange.end) {
-    assert(0);
-  }
+  assert(nextFreeAddr <= lh_memRange.end);
   return curr;
 }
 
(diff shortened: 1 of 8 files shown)


jiamingz9925:
@gc00 I think there is something wrong with dmtcp hook, please verify that

gc00:
I can't update DMTCP.  I'm traveling.  Can you just point DMTCP to before the May 30 commits on your side when you test?

JainTwinkle:
Hi @gc00,

`shmat` requires to unmap the region from the reserved lower half region to fix the following error:
```
sMPICH ERROR [Rank 0] [job id 10705231.0] [Tue Jun 27 08:39:48 2023] [nid006056] - Abort(1091855) (rank 0 in comm 0): Fatal error in PMPI_Init: Other MPI error, error stack:
MPIR_Init_thread(170)...: 
MPID_Init(486)..........: 
MPIDU_Init_shm_init(147): unable to allocate shared memory

aborting job:
Fatal error in PMPI_Init: Other MPI error, error stack:
MPIR_Init_thread(170)...: 
MPID_Init(486)..........: 
MPIDU_Init_shm_init(147): unable to allocate shared memory
```

Please apply the following fix:
```
diff --git a/mpi-proxy-split/lower-half/shmat.c b/mpi-proxy-split/lower-half/shmat.c
index c45fdc89..a9bd8cad 100644
--- a/mpi-proxy-split/lower-half/shmat.c
+++ b/mpi-proxy-split/lower-half/shmat.c
@@ -1,4 +1,6 @@
 #include <sys/shm.h>
+#include <sys/mman.h>
+#include <unistd.h>
 #include <errno.h>
 #include <assert.h>
 #include <string.h>
@@ -33,6 +35,11 @@ __wrap_shmat(int shmid, const void *shmaddr, int shmflg)
       len = ds.shm_segsz;
     }
   }
+  int rc = munmap(addr, len); // munmap some of our reserved memory.
+  if (rc == -1) {
+    char msg[] = "*** Panic: MANA lower half: can't munmap a region for shmat.\n";
+    write(2, msg, sizeof(msg)); assert(rc == 0);
+  }
   void *ret = __real_shmat(shmid, addr, shmflg);
   if (ret != (void*)-1) {
     int idx = getMmapIdx(addr);
```

mpi-proxy-split/lower-half/mmap64.c
```diff
@@ -80,13 +114,49 @@ getMmappedList(int *num)
...
+  // We will reserve the lh_memRange memory, so no one can steal it from us.
+  size_t length = lh_memRange.end - lh_memRange.start;
+#if HAS_MAP_FIXED_NOREPLACE
+  lh_memRange.start = mmap(lh_memRange.start, length,
+             PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS|MAP_FIXED_NOREPLACE, -1, 0);
+  if (lh_memRange.start == MAP_FAILED) {
+    char msg[] = "*** Panic: MANA lower half: can't reserve lh_memRange"
+                 " using MAP_FIXED_NOREPLACE\n";
+    perror("mmap");
+    write(2, msg, sizeof(msg)); assert(rc == 0);
```
jiamingz9925, line 144: I think rc is not defined in this context? @gc00

gc00: Can you fix it directly now and test?
I'll update the PR next.
- Gene

jiamingz9925: I commented it out and am testing the rest of the code, will send email once I gather everything I observed. @gc00

gc00:
Okay.  I've updated the PR to fix the issue of the undeclared 'rc'.  It should compile now.

And I've raised the memory region between lower and upper half from 2GB to 8GB.  I'm not sure if that was there before.  As always, if switching to this branch from another one,, do `git submodule update` and `cd dmtcp && ./configure && make -j`.

jiamingz9925:
Hi Gene,
The issue I described in my last email still exists at latest commit.

Jiaming
________________________________

jiamingz9925 (approved):
LGTM, Thanks Gene.

PR #324, opened 2023-06-22
add FFLAG to test if -fallow-argument-mismatch works for Fortran compiler

gc00:
Use MANA configure to test if `-allow-argument-mismatch`, instead of making use modify `mpi-proxy-split/test/Makefile` each time.

This flag is required for gfortran version 10 to support legacy Fortran code, but it doesn't exist for earlier versions.


dahongli (approved):
Look Good. Thanks Gene!

PR #328, opened 2023-06-29
Fix memory range

dahongli:
The memory layout is changed when no_randomize is enabled. The assumption that lower half would be 1GB below stack region is not valid anymore. This change uses region below minHighMemStart as the lower half.

This change has to used with Gene's DMTCP PR #1063 ADDR_NO_RANDOMIZE.

### DIFF
--- a/restart_plugin/mtcp_split_process.c
+++ b/restart_plugin/mtcp_split_process.c
@@ -289,6 +289,15 @@ setLhMemRange(RestoreInfo *rinfo)
 #if !defined(MANA_USE_LH_FIXED_ADDRESS)
     lh_mem_range.start = (VA)area.addr - TWO_GB;
     lh_mem_range.end = (VA)area.addr - ONE_GB;
+    // High memory region occupies more than 1GB below stack region.
+    // Use minHighMemStart to set lower half memory range.
+    // Plugin_hook reserves region that ends 4GB below high memory region
+    // (See mtcp_plugin_hook). LH reserved 1GB size region that ends at 1GB
+    // below (high memory start - 4GB), so 5GB below high memory.
+    if (lh_mem_range.end > (VA) rinfo->minHighMemStart) {
+      lh_mem_range.end = rinfo->minHighMemStart - 5 * ONE_GB;
+      lh_mem_range.start = lh_mem_range.end - 1 * ONE_GB;
+    }
 #else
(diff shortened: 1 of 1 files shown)


restart_plugin/mtcp_split_process.c
```diff
@@ -289,6 +289,15 @@ setLhMemRange(RestoreInfo *rinfo)
...
     lh_mem_range.end = (VA)area.addr - ONE_GB;
+    // High memory region occupies more than 1GB below stack region.
+    // Use minHighMemStart to set lower half memory range.
+    // Plugin_hook reserves region that ends 4GB below high memory region
+    // (See mtcp_plugin_hook). LH reserved 1GB size region that ends at 1GB
+    // below (high memory start - 4GB), so 5GB below high memory.
+    if (lh_mem_range.end > (VA) rinfo->minHighMemStart) {
+      lh_mem_range.end = rinfo->minHighMemStart - 5 * ONE_GB;
+      lh_mem_range.start = lh_mem_range.end - 1 * ONE_GB;
+    }
```
gc00, line 300: @dahongli ,
    This is a suggestion.  But you've read the code and looked at the /proc/PID/maps more recently than I have.  So, I'm approving the PR now, and you can consider if you want to modify the PR.

QUESTION:  Should we add an MTCP_ASSERT after this, to make clear our _new_ assumption, and quickly catch any future bugs?

In particular, does `rinfo->minHighMemStart` have to be larger than some other address for a memory segment that is lower in the memory layout?  If so, let's specify this directly with an MTCP_ASSERT.

gc00 (approved):

restart_plugin/mtcp_split_process.c
gc00: @dahongli , Thanks for the assert.  This PR looks great.

gc00:
@dahongli,
    I just realized that, either as a commit in this new PR or in a new PR, we need to add the `--mpi` flag to `dmtcp_launch` in `bin/mana_launch`.
On line 182 of  `bin/mana_launch`, we can add it.
We can also add `--mpi` in each of the comments that refers to `dmtcp_launch`.

dahongli:
@gc00 I will file a new PR for adding --mpi to mana_launch. The dashboard automation tests use the dmtcp_launch. I will sync up with Tom to add --mpi flag when using dmtcp_launch.

PR #330, opened 2023-06-30
Add "--mpi" flag to dmtcp_launch in mana_launch

dahongli:
This change uses "--mpi" option to each dmtcp_launch


gc00 (approved):
LGTM.  Thanks.

PR #331, opened 2023-06-30
Use the right flags for building on Perlmutter

dahongli:
This change fixes the broken build on Perlmutter. Please see MANA-310 for the detail of the build failure.


gc00 (approved):
LGTM.  Thanks for this fix.

PR #332, opened 2023-07-06
Add restart_plugin/README (document MANA restart)

gc00:
This `restart_plugin/README` hopefully fixes the lack of documentation on:
1. how DMTCP restart plugins work; and
2. how MANA executes a restart for MPI.


jiamingz9925 (approved):
Looks good, thanks for documenting this!

gc00:
@jiamingz9925 , Thanks for the review.  And please note that I've added an extra sentence at the end, in case it wasn't already obvious:
```
  This can also be checked in GDB by comparing /proc/self/maps inside
    ../dmtcp/src/mtcp/mtcp_restart.c:main() just before it loads the
    checkpoint image file, with the /proc/self/maps when afterward executing
    the statement 'case DMTCP_EVENT_RESTART:' in the file
    ../mpi-proxy-split/mpi_plugin.cpp.
```

PR #333, opened 2023-07-13
WIP: improve-virtualids: O(n) record-replay -> O(1) decode-recode, VID portability.

leonidbelyaev:
This is a WIP PR improving VIDs per Yao Xu's VID proposal.

The essence of it is replacing VID->RID with VID->DESC. DESC in turn contains the RID, and information that *uniquely describes* the RID, such that we can reconstruct it in O(1) time on restart.
VIDs have binary masks / tags that identify their type. Doing away with std::map for an efficient page-table-like structure is a goal.

Points of note are annotated with HACK or TODO.

I maintain the existing macro interface for virtual ids, as it permits flexibility in using different kinds of descriptors (which may have different real-id sizes, and therefore are not interchangeable for the first N elements [MPI_File, notably, is a ptr64]), and means that most existing code in the wrappers is the same. It also permits for casting that depends on types, improving portability for MPI implementations with different type definitions.

Other changes include the addition of a `scripts` directory per Rajat Bisht @rajatpratapbisht , some more documentation for NEU Discovery in `doc/`, and a change from me to `run_all.py`.

Many commits are very minor, as I've been using `git` to ferry code between my machine and Perlmutter. They should probably be squashed when it's time to merge.


gc00:
@leonidbelyaev , Thanks for this critical PR.  I'm looking forward to reorganizing the commits into something shorter.  That helps us in debugging, by testing the first commit, the first two, the first three, etc.

Hopefully, this isn't a large burden.  I would suggest copying this branch to a new branch, tmp1.
Then as you reorganize the commits, remember to copy to tmp2, tmp3, etc.  Finally, every so often, do something like `git diff HEAD tmp1` to make sure you haven't lost any of the code.  Getting familiar with
> git rebase -i HEAD~300

or something, is one way of attacking this.  Or else, start a new branch, with new commits for part 1, part 2, etc., of this large PR (and again do `git diff new_branch this_branch` to make sure you didn't lose anything.

Thanks very much.

leonidbelyaev:
Closing so as to write a better PR.

PR #334, opened 2023-07-16
Updated split-process to use RESERVE_LH_MEM_RANGE

karya0:
This is joint work between @karya0 and @gc00.
* This fixes a 'JWARNING' in 'recordPostMpiInitMaps()' to cast to '*void*)'.
* This fixes PR $323 for mmap64.c:resetMmappedList() to reserve the lh_memRange in advance.  Thsi works now, with the fix to 'JWARNING'.
* A comment was added to restart_plugin/mtcp_split_process.c:initializeLowerHalf() to better explain what happens during restart when a child is fored, exec'ed to execute lh_proxy, copied back into the parent address space, and then initialized in the current (parent) process


gc00:
I assume that this PR is now contained in the dev/karya0/splitProcess branch.  If so, we can close this PR, and eventually approve the PR for dev/karya0/splitProcess.  Does that work?

jiamingz9925 (approved):
LGTM

PR #337, opened 2023-07-26
Use exec at end of mana_launch/restart scripts

gc00:
Using 'exec' at end of scripts for 'ana_launch' and 'mana_restart'.

By doing this, we have one less process.  (The bash script execs into the MANA process.)  So, there are fewer mistakes in identifying the running MANA process.


jiamingz9925 (approved):
LGTM

PR #338, opened 2023-07-26
MPI_Wtime wrapper update

jiamingz9925:
Let MPI_Wtime wrapper directly call wtime() and return the result.

### DIFF
--- a/mpi-proxy-split/mpi-wrappers/mpi_wrappers.cpp
+++ b/mpi-proxy-split/mpi-wrappers/mpi_wrappers.cpp
@@ -108,10 +108,18 @@ USER_DEFINED_WRAPPER(int, Init_thread, (int *) argc, (char ***) argv,
 // FIXME: See the comment in the wrapper function, defined
 // later in the file.
 // DEFINE_FNC(int, Finalize, (void))
-DEFINE_FNC(double, Wtime, (void))
 DEFINE_FNC(int, Finalized, (int *) flag)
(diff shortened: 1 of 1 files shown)


mpi-proxy-split/mpi-wrappers/mpi_wrappers.cpp
```diff
@@ -108,11 +108,19 @@ USER_DEFINED_WRAPPER(int, Init_thread, (int *) argc, (char ***) argv,
 // FIXME: See the comment in the wrapper function, defined
 // later in the file.
 // DEFINE_FNC(int, Finalize, (void))
-DEFINE_FNC(double, Wtime, (void))
 DEFINE_FNC(int, Finalized, (int *) flag)
 DEFINE_FNC(int, Get_processor_name, (char *) name, (int *) resultlen)
 DEFINE_FNC(int, Initialized, (int *) flag)
```
gc00, line 114: I think you need to add:
`#include <time.h>`
at the top of the file (above `#include mpi.h`).

gc00 (commented):
Pls add the '#include <time.h>'.  I suspect that you're seeing a compiler warning right now, because of it.

mpi-proxy-split/mpi-wrappers/mpi_wrappers.cpp
jiamingz9925: Yes, have it added.

gc00 (approved):
Looks good.  Please push it in.  Thanks.

PR #340, opened 2023-08-01
MPI_MANA_Internal: Called in lh on restart (debug)

gc00:
This defines a function, MPI_MANA_internal(char *) that calls the same function in the lower half.  The lower-half function normally does nothing.

This is useful for debugging.  You can keep the checkpoint image, and just modify this function in the lower-half to test something in the lower-half environment.

See the comments in the code:
```
// In mpi-proxy-split/lower-half, redefine MPI_MANA_Internal()
//   to do whatever is desired.  Then do:
//   rm bin/lh_proxy
//   module list # Verify the correct 'module' settings
//   make -j mana
//   bin/mana_coordinator
//   bin/mana_restart
```

### DIFF
--- a/mpi-proxy-split/lower-half/libproxy.c
+++ b/mpi-proxy-split/lower-half/libproxy.c
@@ -46,6 +46,8 @@
 #include "../cartesian.h"
 #endif
 
+extern int MPI_MANA_Internal(char *dummy);
+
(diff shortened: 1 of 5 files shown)


jiamingz9925 (approved):
This debug code block works good for me when testing MANA restart.

PR #341, opened 2023-08-03
Extend restart_plugin/README with more detail

gc00:
This extends `restart_plugin/README` with more details about the sequence of events during MANA restart.  To review it, ignore the github 'diff', and just view the new version of the 'README' file.

@karya0 and @JainTwinkle,
    Could each of you also take a quick look at this README file?
    In addition to suggesting changes, _please do_ suggest additional parts of the restart that should be documented.

One of the reasons that we sometimes have trouble debugging the restart is that even if we once understood the code, we forget it by the next time that we need to debug something.  One of my roles in the collaboration seems to be acting as "secretary"  :-), and updating the documentation.  I'm happy to do that, if this accelerates our debugging sessions.

A second reason for me doing this is that we eventually want to switch from a statically linked lower half to a dynamically linked lower half.  Two students are writing a self-contained prototype based on a  framework provided by @JainTwinkle.  When it comes time to port their prototype to MANA, we will want to review MANA's design, in order to do a clean port from the prototype to MANA.  This  README file should greatly help at that stage, too.

Your careful reviews will help me to ensure that this documentation is easily understood -- even by the two students working on a dynamically linked version of the lower half.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -1,4 +1,3 @@
-@SET_MAKE@
 LN_S=@LN_S@ -f
 MKDIR_P=@MKDIR_P@
 CC=@CC@
--- a/README.md
+++ b/README.md
@@ -3,18 +3,27 @@
 MANA is an implementation of transparent checkpointing for MPI.  It is
 built as a plugin on top of [DMTCP](https://github.com/dmtcp/dmtcp).
 
-We are currently concentrating on making MANA robust (especially on Cori
-and Perlmutter at NERSC).  We plan to roll this ot to other platforms
-(CentOS 7 with MPICH, etc.), over time.  If you have the technical expertise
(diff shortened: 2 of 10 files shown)


gc00:
@jiamingz9925 , @karya0 , @JainTwinkle , @xuyao0127 , @leonidbelyaev :
Hi all.  As I continue to fix low-level MANA issues concerning restart and memory layout, I'm keeping notes on the detailed low-level architecture of MANA for restart.  It's all in a PR to update:
  `restart_plugin/README`
I don't intend to push this in yet, since I'm still noticing new things to document, while I debug.

This documentation may help you in your own struggles with MANA.  :-)  Also, if you have comments on how to improve this documentation, that also is welcome.

PR #343, opened 2023-08-07
Fix libStart/End

gc00:
Still needs testing.

### DIFF
--- a/configure-mana
+++ b/configure-mana
@@ -29,8 +29,7 @@ fi
              MPI_LD_FLAG=-lmpich \
              MPI_CFLAGS= \
              MPI_CXXFLAGS= \
-             MPI_LDFLAGS= \
-             MANA_USE_LH_FIXED_ADDRESS=1
+             MPI_LDFLAGS=
 
 # NOTE:  To make it work, two changes were needed:
 # ===============================================================
--- a/mpi-proxy-split/Makefile_config.in
+++ b/mpi-proxy-split/Makefile_config.in
@@ -1,7 +1,3 @@
-# FIXME: We should eventually dynamically choose this address in setLhMemRange.
-#        Then, we won't need this configure variable.
-MANA_USE_LH_FIXED_ADDRESS = @MANA_USE_LH_FIXED_ADDRESS@
-
 CFLAGS = @CFLAGS@
 CXXFLAGS = @CXXFLAGS@
 FFLAGS = ${CXXFLAGS} @FFLAGS@
--- a/mpi-proxy-split/lower-half/mmap64.c
+++ b/mpi-proxy-split/lower-half/mmap64.c
@@ -21,6 +21,9 @@
 #include <errno.h>
 #include <string.h>
 #include <unistd.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
 #include <sys/mman.h>
 #include <stdio.h>
 #include <assert.h>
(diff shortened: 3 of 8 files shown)


jiamingz9925 (approved):
I tested several workloads and this code can successfully restarted. Just remind that at some point we would like to change how we set minHighMemStart and this is potentially risky.

gc00:
@jiamingz9925 wrote:
> Just remind that at some point we would like to change how we set minHighMemStart and this is potentially risky.

Agreed.  Soon I will push in this PR #343.  Your more risky change to how we set minHighMemStart can be a new PR (and even marked as 'WIP:', if you like.

We can then make your new PR to be the experimental branch, and shift the nightly testing to point to it.

And thanks for adding commit 6fe8a3 ("[assertion updated for == case") to this PR.

jiamingz9925:
> @jiamingz9925 wrote:
> 
> > Just remind that at some point we would like to change how we set minHighMemStart and this is potentially risky.
> 
> Agreed. Soon I will push in this PR #343. Your more risky change to how we set minHighMemStart can be a new PR (and even marked as 'WIP:', if you like.
> 
> We can then make your new PR to be the experimental branch, and shift the nightly testing to point to it.
> 
> And thanks for adding commit 6fe8a3 ("[assertion updated for == case") to this PR.

Hi Gene,
I just want to confirm what do you want from the new PR.
Do you want me separate the current fix and create a new PR? If so are you going to exclude that part from this current PR?

gc00:
@jiamingz9925 ,
You wrote:
>Do you want me separate the current fix and create a new PR? If so are you going to exclude that part from this current PR?

Yes.  Exactly.  This PR #343 now makes MANA reasonably stable, and does not appear to create new bugs.  So, I will soon push PR #343 (the current experimental branch) into main.  You can create a new PR, which will likely become the new experimental branch after some rudimentary testing.

By the way, @leonidbelyaev has discovered two minor bugs in MANA.  One is a potential memory leakage, but is unlikely to affect our common cases.  The second bug has the potential to sometimes cause a segfault on a second restart (i.e., ckpt->restart->ckpt->restart).  It is more likely to occur when there are many ranks.

jiamingz9925:
> @jiamingz9925 , You wrote:
> 
> > Do you want me separate the current fix and create a new PR? If so are you going to exclude that part from this current PR?
> 
> Yes. Exactly. This PR #343 now makes MANA reasonably stable, and does not appear to create new bugs. So, I will soon push PR #343 (the current experimental branch) into main. You can create a new PR, which will likely become the new experimental branch after some rudimentary testing.
> 
> By the way, @leonidbelyaev has discovered two minor bugs in MANA. One is a potential memory leakage, but is unlikely to affect our common cases. The second bug has the potential to sometimes cause a segfault on a second restart (i.e., ckpt->restart->ckpt->restart). It is more likely to occur when there are many ranks.

I think the thing I am concerned is without the temporary fix currently existed in this branch, MANA cannot restart stably. Or maybe I created an experimental PR for properly fix the temporary fix we have currently?

gc00:
@jiamingz9925 ,
You wrote:
> I think the thing I am concerned is without the temporary fix currently existed in this branch, MANA cannot restart stably. Or maybe I created an experimental PR for properly fix the temporary fix we have currently?


Yes, I agree.  So, when I push PR #343 into 'main', it will include the temporary fix that we need for stability.

When you create your new, experimental PR, that will replace the temporary fix by a more risky,, but more robust fix.

PR #345, opened 2023-08-15
Fix PMPI decl. of MPI_Comm_rank (bad orig. decl.)

gc00:
@xuyao0127 and @leonidbelyaev discovered this bug.  The original MANA design for Cray MPI/MPICH didn't care because MPI_Comm and MPI_Group were both int's.  Applying MANA to other MPI's exposed this bug.


xuyao0127 (approved):
LGTM

PR #346, opened 2023-08-15
Warn the end user if no more lower-half memory is available for mmap

gc00:
The default right now is to reserve 1 GB of memory during restart, starting at 0x2aab000... .  Sometimes, we need more than 1 GB.  The error statement tells the end user where to modify MANA, and recompile it with more memory.


leonidbelyaev (approved):
LGTM

PR #347, opened 2023-08-16
Fix compiler error in prev commit; Bad assert

gc00:
Simple one line fix.


leonidbelyaev (changes requested):
Needs to be done on line 206, in the function getNextAddr.

gc00:
@leonidbelyaev , Thanks!  And I found some similar bugs concerning 'assert' with the wrong condition, elsewhere in mmap64.c.  Could you double-check me, and make sure I didn't create a new bug?

Thanks.

leonidbelyaev (approved):

leonidbelyaev:
LGTM now.

PR #349, opened 2023-08-17
Track numMmapRegions ptr instead of copying the value

JainTwinkle:
This is to avoid scenarios where the variable `g_numMmaps` contains stale information of `numRegions` after calling `getLhMmapList` function in the upper half.

### DIFF
--- a/mpi-proxy-split/mpi_plugin.cpp
+++ b/mpi-proxy-split/mpi_plugin.cpp
@@ -87,7 +87,7 @@ bool CheckAndEnableFsGsBase();
 
 EXTERNC pid_t dmtcp_get_real_tid() __attribute((weak));
 
-int g_numMmaps = 0;
+int *g_numMmaps = NULL;
 MmapInfo_t *g_list = NULL;
 mana_state_t mana_state = UNKNOWN_STATE;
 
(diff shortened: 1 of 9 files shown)


mpi-proxy-split/p2p_log_replay.cpp
```diff
@@ -77,16 +77,16 @@ updateCkptDirByRank()
   o << baseDir << "/ckpt_rank_" << g_world_rank;
   dmtcp_set_ckpt_dir(o.str().c_str());
 
-  if (!g_list || g_numMmaps == 0) return;
+  if (!g_list || !g_numMmaps) return;
```
leonidbelyaev, line 80: Is this part correct? Is it possible that g_numMmaps is not null, but points to an int* which does contain 0 segments? 

I guess it might be just checking if this is initialized. If so, one could choose a sentinel value of some kind.

leonidbelyaev:
That's my one contention, looks good to me otherwise.

mpi-proxy-split/mpi_plugin.cpp
```diff
@@ -636,8 +636,8 @@ getLhMmapList()
   if (fnc) {
```
gc00, line 636: I suggest a small comment here, to clarify the key point:
`// Initialize g_numMmaps to point to lower half 'numRegions'`

gc00 (approved):
I have two small suggestions to change (one line each).

JainTwinkle:
@gc00 Thanks! I was also about to push the changes you suggested. The PR looks good. Should I merge now?

gc00:
I'm pushing this in now, so that it will be included in the nightly tests.

PR #351, opened 2023-08-17
lower_half/mmap64.c: mmaps[]: add sentinel

gc00:
This adds a sentinel (MAP_FAILED) to the mmaps[] array, and then checks for it in the mpi_plugin.cpp:DMTCP_EVENT_PRECHECKPOINT.

It also renames numRegions to numMmapRegions in lower_half, and improves the documentation in mmap64.c where we reserve a large mmap'ed region for mmap64.c,, starting at 0x2aab00000000;.

### DIFF
--- a/mpi-proxy-split/lower-half/mmap64.c
+++ b/mpi-proxy-split/lower-half/mmap64.c
@@ -94,7 +94,8 @@
 //  2. Make the size of list dynamic
 
 // Number of valid objects in the 'mmaps' list below
-int numRegions = 0;
+// For i >= numMmapRegions, mmaps[i].addr == MAP_FAILED.
+int numMmapRegions = 0;
 
 #ifdef LIBMMAP_SO
 // Lower-half memory range to use.
@@ -113,7 +114,7 @@ MmapInfo_t*
 getMmappedList(int **num)
 {
   if (!num) return NULL;
-  *num = &numRegions;
+  *num = &numMmapRegions;
   return mmaps;
 }
 
@@ -130,10 +131,14 @@ void
 resetMmappedList()
 {
   // Alternatively, just do:  memset((void *)mmaps, 0, sizeof(mmaps));
-  for (int i = 0; i < numRegions; i++) {
+  for (int i = 0; i < numMmapRegions; i++) {
     memset(&mmaps[i], 0, sizeof(mmaps[i]));
   }
-  numRegions = 0;
+  numMmapRegions = 0;
+  // All mmaps[] now unused.  Mark all mmaps with MAP_FAILED sentinel, for GDB
+  for (int i = 0; i < MAX_TRACK; i++) {
+    mmaps[i].addr = MAP_FAILED;
+  }
 
   /**************************************************************************
    * Next, We reserve the lh_memRange memory, so no one can steal it from us.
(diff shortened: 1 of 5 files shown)


mpi-proxy-split/lower-half/mmap64.c
```diff
@@ -119,7 +114,7 @@ MmapInfo_t*
 getMmappedList(int **num)
 {
   if (!num) return NULL;
-  *num = &numRegions;
```
leonidbelyaev, line 122: Am I missing something, or does this undo the pointer-tracking work?

JainTwinkle: I agree! This is not correct. `*num` should hold the address of the `numMmapRegions` variable.

mpi-proxy-split/lower-half/mmap64.c
```diff
@@ -136,27 +131,25 @@ void
 resetMmappedList()
 {
   // Alternatively, just do:  memset((void *)mmaps, 0, sizeof(mmaps));
-  for (int i = 0; i < numRegions; i++) {
+  for (int i = 0; i < numMmapRegions; i++) {
     memset(&mmaps[i], 0, sizeof(mmaps[i]));
+    mmaps[i].addr = MAP_FAILED; // Mark those mmaps past numMmapRegions, for GDB
```
JainTwinkle, line 136: memset should take care of setting all the variables to 0x0. We do not need to set MAP_FAILED separately. Also, the comment does not reflect the code. The loop is only till numMmapRegions, not past that.

mpi-proxy-split/lower-half/mmap64.c
```diff
@@ -401,28 +393,36 @@ __mmap64 (void *addr, size_t len, int prot, int flags, int fd, __off_t offset)
...
+        //   LH_MMAP_CALL(lh_memRange.end, additional_length, PROT_NONE,
+        //                MAP_SHARED|MAP_ANONYMOUS|MAP_FIXED_NOREPLACE, fd, 0)
+        //   for fd pointing to /dev/zero, and so to contiguously extend
+        //   our reserved mmaps[] region, and then adjust MAX_TRACK.
+        if (numMmapRegions >= MAX_TRACK - 2) { // Keep a MAP_FAILED sentinel
           char msg[] = "*** Panic: MANA lower half: no more space for mmap.\n";
-          write(2, msg, sizeof(msg)); assert(numRegions > 0);
+          write(2, msg, sizeof(msg));
         }
+        assert(numMmapRegions < MAX_TRACK - 2); //MAX_TRACK-2 in case guard page
```
JainTwinkle, line 419: Minor: space after //

mpi-proxy-split/lower-half/mmap64.c
JainTwinkle: Did you mean to add MAP_FAILED after the for loop? The value of `i` would then be `numMmapRegions`.

mpi-proxy-split/lower-half/mmap64.c
gc00: Thanks!  You're right.  I forgot the '&'.  I'll fix that.

mpi-proxy-split/lower-half/mmap64.c
gc00: Done.

gc00:
@JainTwinkle, Thanks for the comment about:
> Did you mean to add MAP_FAILED after the for loop? The value of i would then be numMmapRegions.

Good catch!    I've now changed it to set the 'addr' to MAP_FAILED for _all_ mmaps elements.  Since 'numMmapRegions' is 0 at the end, I could set it just for the zero-th element.  But it's safer to set it for all 'mmaps' elements.

PR #353, opened 2023-08-28
Don't skip munmap of mtcp_restart regions.

karya0:


restart_plugin/mtcp_restart_plugin.c
```diff
@@ -849,6 +849,11 @@ mtcp_plugin_skip_memory_region_munmap(Area *area, RestoreInfo *rinfo)
   LhCoreRegions_t *lh_regions_list = NULL;
   int total_lh_regions = lh_info->numCoreRegions;
 
+  // Don't skip munmap of mtcp_restart regions.
```
gc00, line 852: Let's change the comment to remove the double negative:
`// Do an munmap of mtcp_restart regions during restart.  Don't skip this.`

Also, please add a comment about why we need to munmap the mtcp_restart regions within MANA, but we don't need to do that within ordinary DMTCP.  Where is the potential address conflict that we're trying to avoid?

gc00 (changes requested):
Please modify the comment, and add the extra requested comment.  Otherwise, in the future, we look back at this mysterious special case, and wonder why.

Separately, I assume that you're going to squash the two commits together, before pushing this in.

I'd like to wait to see the added comment before approving, just to make sure we're documenting the code well.  Thanks.

restart_plugin/mtcp_restart_plugin.c
jiamingz9925: I want to add that we should not skip all [heap] but only the [heap] right after the mtcp_restart region.

restart_plugin/mtcp_restart_plugin.c
```diff
@@ -849,6 +849,12 @@ mtcp_plugin_skip_memory_region_munmap(Area *area, RestoreInfo *rinfo)
...
+  if (mtcp_strendswith(area->name, "/mtcp_restart") ||
+      mtcp_strendswith(area->name, "[heap]")) {
```
jiamingz9925, line 854: we only want to skip the [heap] right after mtcp_restart. Also in mpi_plugin.cpp we need to do the same to skip those area for libsStart consideration.

jiamingz9925, line 854: Update: For some reason unmap the heap region here would cause seg fault, but without unmapping it we will encounter conflict as well later. @karya0

gc00:
@karya0 ,
    This issue is a blocker for MANA development.  I hope you can get back to it soon.
Best,
- Gene

jiamingz9925:
@karya0 @gc00 I can try to do some experiment in my forked repo and based on this PR as well

gc00:
See PR #357 for the continuation of this analysis.  We should probably close this PR without committing

gc00:
@karya0 , If this PR #353 is now obsolete, can you close it?

PR #354, opened 2023-08-28
Simplify copy bits

gc00:
This has various bug fixes from dev/gdc0/simplifyCopyBits on origin, as well as cleaning up some code.  I've tried to re-organize the order of the commits so that the obviously important commits come first.  There was a report by @JainTwinkle that some of the commits caused problems when experimenting with Open MPI.  I'm hoping that at least up to 31c953882529e6c21 , everything is vanilla, and should work.

@JainTwinkle , could you test up to this commit (any maybe to the last commit), and tell us what works and what doesn't work for Open MPI?

Thanks.

### DIFF
--- a/mpi-proxy-split/mpi_plugin.cpp
+++ b/mpi-proxy-split/mpi_plugin.cpp
@@ -699,7 +699,9 @@ computeUnionOfCkptImageAddresses()
   void *libsEnd = NULL;
   void *maxLibsEnd = NULL;
   void *highMemStart = 0x0;
-  void *minHighMemStart = NULL;
+  void *minHighMemStart = 0x0;
+  void *highMemEnd = 0x0; // FIXME: For now, always 8 MB above highMemStart
+  void *maxHighMemEnd = 0x0;
   void *lhEnd = 0x0; // lower bound for actual lhEnd
   void *minAddrBeyondHeap = NULL;
   void *maxAddrBeyondHeap = NULL;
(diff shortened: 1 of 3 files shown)


gc00:
@JainTwinkle ,
    Thanks again for agreeing to test this PR against MANA/Open MPI.
    If you want to simply do something quick, I suggest testing with _only_ the commits through:
    3608f94749ea9ca40c44a0fa0cb535a9fc9c07d0
    "[TODO: split_process.cpp: MANA_USE_LH_FIXED_ADDRES]"

This is the same as:
```
  git fetch gc00 simplifyCopyBits:simplifyCopyBits
  git checkout simplifyCopyBits
  # Next, ignore the last two commits for your testing
  git checkout simplifyCopyBits^^
```

Please let me know if this works with MANA/Open MPI, and if not, then how many of the earlier commits will work.  Then I will push into 'main' whichever commits seem to work with MANA/Open MPI.
    Most of this code is a few very simple fixes, followed by massive cleaning up of the code that should still retain the original semantics.
 
As I explained, I'm trying to clean up the ugly code from the academic prototype.  Then I'll push the cleaner code into 'main'.
    The alternative would be for me to add on top of the existing ugly code, and then later port it to the clean code.  But it will make my life a lot easier if you can test the clean code now, and then I can add on top of the clean code directly.  Thanks!!

gc00:
@JainTwinkle ,
    Could you please test this PR against MANA/Open MPI one more time?

    Thanks for catching my earlier bug with where to save lh_info.  I believe I've fixed it now.  I'm only using lh_info_addr (LowerHalfInfo_t *), and not lh_info itself.  So, it's a pointer into the lh_proxy memory.

    If you want to simply do something quick, I suggest testing with _only_ the commits through:
    24bd2120ef77b1f1ce4e53cd7a0f5454bf47c8e5
    "[TODO: split_process.cpp: MANA_USE_LH_FIXED_ADDRES]"

This is the same as:
```
  git fetch gc00 simplifyCopyBits:simplifyCopyBits
  git checkout simplifyCopyBits
  # Next, ignore the last two commits for your testing
  git checkout simplifyCopyBits^^
```

Please let me know if this works with MANA/Open MPI, and if not, then how many of the earlier commits will work.  Then I will push into 'main' whichever commits seem to work with MANA/Open MPI.
    As before, each individual commit is small, and the goal for most of the commits is to keep the same semantics, but make the code cleaner.

For some reason, I don't see these bugs when testing against wave_mpi.  So, your test with MANA/OpenMPI is very valuable.  Thanks!

gc00:
@JainTwinkle ,
    I just force-pushed to PR #354 to fix the issue with the dmtcp submodule that you observed.  The DMTCP submodule should now be updated only with the corresponding commit, and you're testing only before that commit.

The code to test is exactly the same.
When you do `git checkout simplifyCopyBits^^`, the code will be the same, but the hash for that commit will have changed.

You can continue to work with your old version, or do a `git pull --rebase` on this updated PR.

Thanks again for this testing!!!

gc00:
This has two remaining commits that may or may not be useful.  Most of the original commits disappeared when we pushed a lot of dev/gdc0/simplifyCopyBits into main.

I'll come back later and see if these commits are useful.

PR #358, opened 2023-09-06
bin/mana_*: Use $HOME/.mana-slurm-$SLURM_JOB_ID.rc

gc00:
This adds the SLURM_JOB_ID to the bin/mana_* scripts.  This allows multiple SLURM jobs to run safely.

Otherwise, two jobs could both call `bin/mana_coordinator` simultaneously, overwriting the `.mana` file with a single file.  This later causes two independent jobs to use a single `.mana` file when running `bin/mana_launch` or `bin/mana_restart`.

### DIFF
--- a/bin/mana_launch
+++ b/bin/mana_launch
@@ -47,6 +47,10 @@ while [ -n "$1" ]; do
     options="$options $1 $2"
     shift
   else
+    if [ "$1" == --quiet ] || [ "$!" == "-q" ]; then
+      export MANA_QUIET=1
+      # And --quiet is also a flag for DMTCP. Continue from here.
+    fi
     # other flags, options, and target_app executable
     options="$options $1"
     # The last word will be the target_app.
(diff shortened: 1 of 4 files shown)


bin/mana_launch
```diff
@@ -193,7 +202,7 @@ fi
 #   needs to send/receive a large message.
 
 exec env LD_LIBRARY_PATH="$libdir:$LD_LIBRARY_PATH" \
-    $dir/dmtcp_launch --coord-host $submissionHost \
+    $dir/dmtcp_launch ---mpi --coord-host $submissionHost \
```
leonidbelyaev, line 205: three hyphens on `mpi`?

leonidbelyaev (approved):

leonidbelyaev (approved):
Looks good, but see my comment about three hyphens.

JainTwinkle:
@gc00,  Can the coordinator put the same unique string as a suffix to the `.mana` file as DMTCP add for the checkpoint images?

gc00:
@leonidbelyaev , Thanks!  (Doing too many things at once.  :-(  )
It's fixed now.

gc00:
@JainTwinkle wrote:
> Can the coordinator put the same unique string as a suffix to the .mana file that DMTCP adds to the checkpoint images?

I don't think that works.  The coordinator creates the .mana file first.  And then we call mana_launch, which reads it.  There is no ckpt image yet.  But maybe I misunderstood your comment.

JainTwinkle:
> @JainTwinkle wrote:
> 
> > Can the coordinator put the same unique string as a suffix to the .mana file that DMTCP adds to the checkpoint images?
> 
> I don't think that works. The coordinator creates the .mana file first. And then we call mana_launch, which reads it. There is no ckpt image yet. But maybe I misunderstood your comment.

I see your point. The coordinator is started first and does not have information about the unique string that will be a part of the checkpoint image's name in the future. In that case, your current design is fine.

gc00:
It says that I closed this PR on Sept. 6.  If so, it was an accident.  I'm re-opening it now.

PR #361, opened 2023-09-11
Clean up lower half and restart logic

gc00:
These commits have been part of the experimental branch, `dev/gdc0/simplifyCopyBits`, with no bugs reported recently.  Let's push this into main.  We'll still keep  `dev/gdc0/simplifyCopyBits` as the experimental branch, and we can do a `git pull --rebase` of main to bring it up to date.


jiamingz9925 (approved):

jiamingz9925 (changes requested):
As @JainTwinkle suggested, we might need to do a pull from main to have the commit messages squashed.

restart_plugin/mtcp_restart_plugin.c
```diff
@@ -594,21 +602,28 @@ mtcp_plugin_hook(RestoreInfo *rinfo)
...
     // Ignore region start2:end2 if it is overlapped with region start1:end1
     if (is_overlap(start1, end1, start2, end2)) {
       start2 = 0;
       end2 = 0;
     }
   }
-  // FIXME:  End of '#if 1'; Remove '# else' branch when the code is stable.
+
+  Area vvar_area;
+  MTCP_ASSERT(getMappedArea(&vvar_area, "[vvar]") == 1);
```
JainTwinkle, line 621: @gc00

Let's either improve this code to call getMappedArea for vvar only if vvar actually exists in the proc maps, or else enclose it in # if ... endif by checking kernel version. This way a single codebase can work on both Perlmutter and the discovery cluster.

restart_plugin/mtcp_restart_plugin.c
```diff
@@ -719,17 +734,23 @@ mtcp_plugin_hook(RestoreInfo *rinfo)
...
+    end2 = rinfo->maxHighMemEnd;
     // Ignore region start2:end2 if it is overlapped with region start1:end1
     if (is_overlap(start1, end1, start2, end2)) {
       if (end1 < end2) { end1 = end2; }
       start2 = 0;
       end2 = 0;
     }
 
+    Area vvar_area;
+    MTCP_ASSERT(getMappedArea(&vvar_area, "[vvar]") == 1);
```
JainTwinkle, line 749: @gc00  see my previous comment on the same function call.

restart_plugin/mtcp_split_process.c
```diff
@@ -69,23 +78,27 @@ splitProcess(RestoreInfo *rinfo)
...
   pid_t childpid = startProxy(rinfo);
   int ret = -1;
   if (childpid > 0) {
     // Parent has read from pipefd_out; child is ready.
     // We then "copy the bits" of the child process into our (parent's) space.
+    // And it is then safe to copy the lh_info from the child proxy to uh.
     // We then kill the child process.
     ret = read_lh_proxy_bits(rinfo, childpid, rinfo->argv[0]);
+    // Populate lh_info now that we have copied the bits.
+    mtcp_memcpy(&lh_info, lh_info_addr, sizeof(lh_info));
```
JainTwinkle, line 94: @gc00 
Though this design works fine and there are multiple ways to do a single task, why do we need to populate the whole `lh_info` structure? One can simply use the `lh_info` structure directly from the `lh_proxy's` data section. The `lh_info's` address in the lower half is valid and accessible directly (without needing to switch context) throughout the life of the process.

gc00:
@JainTwinkle , Thanks for the great comments.  First, I pushd the wrong branch into this PR.  Too many branches, with probles of version control.  That's why I want to push the "good commits" into 'main'.

In this version, we no longer have:
```
// Populate lh_info now that we have copied the bits.
    mtcp_memcpy(&lh_info, lh_info_addr, sizeof(lh_info));
```

Instead, we have:
```
// FIXME:  We should use lh_info_addr, in place of rinfo->pluginInfo
mtcp_memcpy(&rinfo->pluginInfo, lh_info_addr, sizeof(*lh_info_addr)); 
```

But your second point is well taken.  Right now, the code says:
```
  uint64_t prev_addr = 0x10000;
   ...

  while (mtcp_readmapsline(mapsfd, &area)) {
    if (prev_addr + vvarSize + vdsoSize <= (uint64_t) area.addr) {
      vvarStartTmp = prev_addr;
      vdsoStartTmp = prev_addr + vvarSize;
      break;
    } else {
      prev_addr = ROUNDADDRUP((uint64_t) area.endAddr, MTCP_PAGE_SIZE);
    }
  }
```

I will change it to test if there is a vvar segment.  And then I'll push in a newer version for review.

gc00:
@JainTwinkle, I've added some extra commits for better compatibility with CentOS.  Please try this out on the Discovery cluster.

However, the `lower-half/README.mpich-static` file is still missing something.  When I do `make` in `lower-half`, I can set my LD_LIBRARY_PATH to include the libraries from the README:   `libxml2.a`, `libzma.a`, and `libz.a`.
But when I try this and compile the lower-half, I'm still missing some undefined functions from:
/shared/centos7/ucx/src/ucx-1.10.1
What is the extra library related to ucx that we should tell people to include when building lh_proxy?

gc00:
@JainTwinkle ,
    @karya0 asked if the review of this PR will be completed soon, so that we can push it in.  He wants to build on this base code.  Will you have a review ready soon?  Thanks.

JainTwinkle:
Hi @gc00 

I tried this PR on Discovery cluster today. You'll need one minor change to this PR:
```
-  // update vDSO linkmap entry to the temporary address
-  updateVdsoLinkmapEntry(rinfo->currentVdsoStart,
-                         lh_info_addr->vdsoLdAddrInLinkMap);
+  // update vDSO linkmap entry (if exists) to the temporary address
+  if (lh_info_addr->vdsoLdAddrInLinkMap) {
+    updateVdsoLinkmapEntry(rinfo->currentVdsoStart,
+                           lh_info_addr->vdsoLdAddrInLinkMap);
+  }

```
Checkpoint-restart of ping-pong test case works on discovery after applying the fix.

JainTwinkle (approved):
I've tested this PR on both Perlmutter and Discovery. A single checkpoint-restart works for at least ping pong (2 ranks) on both platforms. Please address the comments. I'm approving this PR in advance.

jiamingz9925 (approved):

gc00:
@JainTwinkle ,
    I've made the fix that you suggest.  And I did a sanity test on Perlmutter, and the code there is still fine.

QUESTION:  I thought that on kernel 3.10 on Discovery, the '[vdso]' area is always labelled, and I thought that this would imply that `lh_info_addr->vdsoLdAddrInLinkMap` always exists.  So, why is this pointer NULL on Discovery?

(And I'll fix up the two comments as you suggest, next.)

JainTwinkle:
> @JainTwinkle , I've made the fix that you suggest. And I did a sanity test on Perlmutter, and the code there is still fine.
> 
> QUESTION: I thought that on kernel 3.10 on Discovery, the '[vdso]' area is always labelled, and I thought that this would imply that `lh_info_addr->vdsoLdAddrInLinkMap` always exists. So, why is this pointer NULL on Discovery?
> 
> (And I'll fix up the two comments as you suggest, next.)

@gc00  It's not about the label. It's about having an entry in the link map. The `lh_proxy` on the Cori platform didn't need this extra step after copying the lower-half bits either. Therefore, I believe that older kernel versions didn't require link maps to have an entry for vdso whereas the newer kernels like Perlmutter do require it.

gc00:
@JainTwinkle, Thanks for that explanation.  I've expanded the comment around ``lh_info_addr->vdsoLdAddrInLinkMap` .

@karya0, I'm pushing this PR #361 into 'main' now.  It's ready to build upon.

PR #363, opened 2023-09-13
REVERT potential resource leak fix (and added new comments)

leonidbelyaev:
In PR #348 , I wrote code that would remove the MPI_Group `group_world` created in `registerLocalSendsAndRecvs` unconditionally.

But this is problematic because of the semantics of the ADD_NEW macro. See virtual-ids.h, lines 214-223. If the group does not yet exist, this is indeed an extra group. However, if the group already exists (i.e., the application has created the same `group_world`), then there is a collision, and the free can make a restart fail, for it erases a group that the application depends on, or one of the child groups depends on. See `mpi-proxy-split/mpi-wrappers/mpi_group_wrappers`, line 100-105. (This is in some sense fixed by virtual-ids, in that there is no more dependency between groups in this way).

This can cause an extra LOG_CALL and LOG_REPLAY to occur as written, but it is not a huge problem. We are working to remove LOG_CALL and LOG_REPLAY altogether in `virtualids-v2` anyways.

For now, we should probably revert these changes for the main branch. I have done so, and also wrote some commentary to explain the situation.

### DIFF
--- a/mpi-proxy-split/p2p_drain_send_recv.cpp
+++ b/mpi-proxy-split/p2p_drain_send_recv.cpp
@@ -45,7 +45,6 @@ extern int MPI_Comm_create_group_internal(MPI_Comm comm, MPI_Group group,
 extern int MPI_Comm_free_internal(MPI_Comm *comm);
 extern int MPI_Comm_group_internal(MPI_Comm comm, MPI_Group *group);
 extern int MPI_Group_free_internal(MPI_Group *group);
-extern int MPI_Comm_internal_virt_group(MPI_Comm comm, MPI_Group *group);
 int *g_sendBytesByRank; // Number of bytes sent to other ranks
 int *g_rsendBytesByRank; // Number of bytes sent to other ranks by MPI_rsend
 int *g_bytesSentToUsByRank; // Number of bytes other ranks sent to us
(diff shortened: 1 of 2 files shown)


mpi-proxy-split/p2p_drain_send_recv.cpp
```diff
@@ -76,7 +75,15 @@ registerLocalSendsAndRecvs()
   // Get a copy of MPI_COMM_WORLD
   MPI_Group group_world;
   MPI_Comm mana_comm;
-  MPI_Comm_internal_virt_group(MPI_COMM_WORLD, &group_world);
+  // MPI_Comm_group creates the same real id (and therefore, virtual id)
+  // for identical calls. If the application already has this virtual id, no
+  // extra resource is created.
+  //
+  // See mpi-proxy-split/virtual-ids.h, 214-223.
```
gc00, line 82: Line numbers like `214-223` can change over time.  Can you identify this by the function or declaration that it relates to in mpi-proxy-split/virtual-ids.h?

gc00 (approved):
Can you please see my request to update your comment, and do that?

jiamingz9925:
@gc00 @leonidbelyaev Hi all, do you mind waiting for me to do a thorough test on VASP before pushing the code in?

leonidbelyaev:
@jiamingz9925 I do not mind. I think that Yao said that this does not reverse the hanging issue you are also facing, though.

jiamingz9925:
> @jiamingz9925 I do not mind. I think that Yao said that this does not reverse the hanging issue you are also facing, though.

I think I knew some tricks to work around that, mainly want to make sure I can restart multiple times with VASP/PdO4 case.

jiamingz9925:
I did not see any problem during 4 nodes 128 ranks run. I was able to restart multiple times

jiamingz9925 (approved):

PR #366, opened 2023-09-14
Simplified mtcp-restart plugin interface.

karya0:
This uses the newly proposed environment variable-based mechanism. The simplified interface also helps decouples DMTCP/MANA even further. 

Corresponding DMTCP PR: https://github.com/dmtcp/dmtcp/pull/1078

### DIFF
--- a/restart_plugin/dmtcp_restart_plugin.cpp
+++ b/restart_plugin/dmtcp_restart_plugin.cpp
@@ -1,6 +1,7 @@
 #include "workerstate.h"
 #include "dmtcp_restart.h"
 #include "jassert.h"
+#include "jconvert.h"
 #include "jfilesystem.h"
 #include "util.h"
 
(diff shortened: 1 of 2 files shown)


restart_plugin/dmtcp_restart_plugin.cpp
```diff
@@ -39,30 +40,22 @@ void dmtcp_restart_plugin(const string &restartDir,
...
-
-  mtcpArgs.push_back((char*) "--minLibsStart");
-  mtcpArgs.push_back((char*) kvmap.at("MANA_MinLibsStart").c_str());
-
-  mtcpArgs.push_back((char*) "--maxLibsEnd");
-  mtcpArgs.push_back((char*) kvmap.at("MANA_MaxLibsEnd").c_str());
-
-  mtcpArgs.push_back((char*) "--minHighMemStart");
-  mtcpArgs.push_back((char*) kvmap.at("MANA_MinHighMemStart").c_str());
+  vector<string> mtcpEnvironment = getMtcpEnvironment(t);
```
gc00, line 43: I see that `getMtcpEnvironment()` is defined in PR #1078 of DMTCP (https://github.com/dmtcp/dmtcp/pull/1078).  As commented there, I think this is too much non-local design.
    Since mtcp_restart is going to throw away any extraneous environment variables when it restores the checkpoint image, let's just pass the _entire_ environment to 'mtcp_restart' through the global `environ` variable.   'mtcp_restart' is going to add special handling for the special environment variables in the 'mtcp' directory.  There's no reason to spread that logic to `restart_plugin/mtcp_restart_plugin.cpp` and to `DMTCP:src/dmtcp_restart.cpp`.
    In the worst case, if you feel that `environ` is not suitable for some reason (please document with a comment), then we should just define `getEnvironment()`.  But I think `environ` should suffice.
    Separately, I assume that this PR will add a commit to include the latest DMTCP, which depends on first finishing PR #1078 of the DMTCP repo.

gc00 (changes requested):
I like this PR a lot, and I'm eager to push it in, and simplify the design of MANA.

But please see my comment about the issue of non-local program design.  Let's just pass in the entire environment.  Thanks.

gc00 (approved):
@karya0 , This looks great.  Thanks.

PR #371, opened 2023-09-18
[REDONE]: Simplified mtcp-restart plugin interface.

karya0:
This uses the newly proposed environment variable-based mechanism.

### DIFF
--- a/bin/mana_launch
+++ b/bin/mana_launch
@@ -214,6 +214,11 @@ find $HOME/.mana*.rc -ignore_readdir_race -maxdepth 0 -mtime +7 -type f -delete
 #   The /dev/xpmem segments are created on demand when an MPI call
 #   needs to send/receive a large message.
 
+# FIXME:  This is temporary, for debugging by developers.  We should add this logic
+#         instead to configure-mana (which should be renamed to configure-discovery).
+#         Use this commit for testing, but don't push it into the main branch.
+export MANA_DEBUG=1
+
 exec env LD_LIBRARY_PATH="$libdir:$LD_LIBRARY_PATH" \
     $dir/dmtcp_launch --mpi --coord-host $submissionHost \
           --coord-port $submissionPort --no-gzip \


gc00 (commented):
@karya0 , Thanks for the quick response.
The two fixes of typos definitely help.  But I'm still seeing the same bug on restart:
    gdc0@perlmutter:login34:~/mana.git> bin/mana_restart
    [2266209] /global/u2/g/gdc0/mana.git/restart_plugin/mtcp_restart_plugin.c:201 getCkptImageByDir:
    ***ERROR No restart directory found - cannot find checkpoint image by directory!Segmentation fault   

Separately, I suspect there's a problem with my DMTCP commit when joining:
  Set "stale timeout" (secs), when no peer processes

However, I can revert that in DMTCP, and I still see the segfault for mana_restart.
Roughly, at checkpoint time, it's too early to record the relative path to the checkpoint directory.  During restart, we ,might be in a different directory.

So, probably bin/mana_restart needs to set the environment variable MANA_RestartDir.  I'm trying to test something there, but it's not yet working for me.

gc00:
I am ensuring that MANA_RestartDir is being set.

I'm looking right now at:
> setenv(key.c_str(), ckptImages[i].c_str(), 1);

in dmtcp_restart_plugin.cpp

I think that the problem is there.  But I'm still testing.

gc00:
Yep.  That's the root cause.  In:
> for (size_t i = 0; i < ckptImages.size(); i++)

`I'm seeing: ckptImages.size() == 0`

gc00:
@karya0 ,
    Okay.  I fixed some more bugs with the original PR.  I pushed this in, as a second commit.

On the DMTCP side, I've also reverted the `"Set "stale timeout" (secs), when no peer processes` commit.  I needed to revert that commit, or else we don't get a checkpoint image after launching using `mana_launch -i5 ...`  (And if you revert it, don't do `git submodule update` after that, or it will undo the revert.)

On that basis, the code now seems to work.  But I will revert the dev/gdc0/simplifyCopyBits branch that we're using for testing.  This code in this PR is still too new

gc00:
Jenkins reports:
```
15:15:17 + git submodule update --init
15:15:17 Submodule 'dmtcp' (https://github.com/dmtcp/dmtcp) registered for path 'dmtcp'
15:15:17 Cloning into 'dmtcp'...
15:15:22 fatal: reference is not a tree: afc5b3c78594f0f12ece4b65d5e6eeb65f8591a0
15:15:22 Unable to checkout 'afc5b3c78594f0f12ece4b65d5e6eeb65f8591a0' in submodule path 'dmtcp'
15:15:22 + ./configure
```
Which branch is this using for DMTCP?  Could it be that it's using a branch that's not part of the origin repo for DMTCP?

PR #373, opened 2023-09-23
Custom dmtcp_restart_MANA and mtcp_restart_MANA binaries.

karya0:
This eliminates the need to modify dmtcp_restart/mtcp_restart files/binaries. The corresponding DMTCP refactor is proposed in https://github.com/dmtcp/dmtcp/pull/1083 and https://github.com/dmtcp/dmtcp/pull/1084.

__There is also PR #1085 from DMTCP.  Should this be included in this PR? - @gc00_

### DIFF
--- a/mpi-proxy-split/mpi_plugin.cpp
+++ b/mpi-proxy-split/mpi_plugin.cpp
@@ -74,6 +74,7 @@ extern CartesianProperties g_cartesian_properties;
 extern ManaHeader g_mana_header;
 extern std::unordered_map<MPI_File, OpenFileParameters> g_params_map;
 
+constexpr uint64_t RestoreBufLen = 32 * 1024 * 1024;
 constexpr const char *MANA_FILE_REGEX_ENV = "MANA_FILE_REGEX";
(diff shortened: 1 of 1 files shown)


gc00:
@karya0 ,  Your PR is not building.  Jenkins reports:
```
02:17:15 configure: creating ./config.status
02:17:15 config.status: creating Makefile
02:17:15 config.status: creating manpages/Makefile
02:17:15 config.status: WARNING:  'manpages/Makefile.in' seems to ignore the --datarootdir setting
02:17:15 config.status: creating mpi-proxy-split/Makefile_config
02:17:15 config.status: error: cannot find input file: `restart_plugin/Makefile.in'
```

gc00 (approved):
Sorry it took me so long.

We don't need it for this PR, but we _really_ should add more documentation about how to use the restart plugins.  This would have also helped me to review this with more confidence.

Please note that we already have a file:
  `README.restart_plugin`  (at the root directory of MANA)

Can we use that as a starting point, and then update the documentation?

@karya0 , if you can do a first draft and then create a PR, I'd be happy to do the second draft.

PR #375, opened 2023-10-03
Add libnl.a to build MANA

dahongli:
This change adds libnl-3.a library to build lower-half. It also fixes the lower half makefile to use the right version of libpal.a on Perlmutter.


gc00 (approved):
@dahongli , Icm approving this in advvance.  But Can you also add to the Makefile in lower-half some statement like:
```
# NOTE:  The static library libnl03.a must be built specially.  A suggested recipe is:
# Search for and download libnl3-3.3.0-1.29.src.rpm
# rpm -i libnl3-3.3.0-1.29.src.rpm
# rpmbuild -ba ~/rpmbuild/SPECS/libnl3.spec
# cd ~/rpmbuild/BUILD/libnl-3.3.0/lib/.libs/
# Build libnl-3.a from lib_libnl_3_la-*.o
```
Thank you for the quick work.

PR #376, opened 2023-10-06
Fix mulitple definitions of json calls

dahongli:
This change solves the multiple definitions problem between libjson-c.a and libjansson.c


mpi-proxy-split/lower-half/static_libs.txt
```diff
@@ -1 +1 @@
--lpmi -lpmi2 /opt/cray/libfabric/1.15.2.0/lib64/libfabric.a -lcxi -L/global/homes/z/zz217/local/lib -L/global/cfs/cdirs/cr/pm_dependencies -ljson-c -lcurl  -lssl -lcrypto -lz -latomic -lpthread /opt/cray/pe/pals/1.2.11/lib/libpals.a libnl-3.a
+-lpmi -lpmi2 /opt/cray/libfabric/1.15.2.0/lib64/libfabric.a -lcxi -L/global/homes/z/zz217/local/lib -L/global/cfs/cdirs/cr/pm_dependencies -ljson-c -lcurl  -lssl -lcrypto -lz -latomic -lpthread /opt/cray/pe/pals/1.2.11/lib/libpals.a libnl-3.a -ljansson -Xlinker -zmuldefs
```
gc00, line 1: Please add to the Makefile near static_libs.txt a comment:
     `# -zmuldefs used in static_libs.txt because libjansson.a and libjson.a both want to define [FILL IN}.`
     `# We prefer libjansson.a`

gc00 (approved):
I'm approving this in advance.  Please add the suggested comment (or something similar) to the Makefile.  In particular, please name the symbol that is multiply deffied

I'm approving this in advance, but please make the suggeted change.

mpi-proxy-split/lower-half/static_libs.txt
dahongli: Good point. Done. Thanks Gene!

PR #378, opened 2023-10-13
Fix DMTCP so that 'dmtcp_coordinator -iXXX' works

gc00:
We need to push DMTCP PR #1090 in, before committing this PR, since this should be changed to refer to a specific DMTCP commit on the 'master' branch.


gc00:
This commit should not be pushed in.  On the DMTCP side, it will be replaced by:
  a new version of the function resetCkptTimer() (see comments) in dmtcp_coordinator.cpp in the
  dev/gc00/betterCoordStaleTimeout  branch.

After that, MANA will eventually inherit the latest 'master' branch from DMTCP.

PR #379, opened 2023-10-29
Fixed errors related to lower-half memory region book-keeping.

karya0:


mpi-proxy-split/lower-half/munmap.c
```diff
@@ -108,13 +93,23 @@ removeMappings(void *addr, size_t len)
...
   if (rc == 0) {
     removeMappings(addr, len);
   }
   return rc;
 }
 
+int
+munmap(void *addr, size_t len)
+{
+  if (addr < lh_info.memRange.start || addr >= lh_info.memRange.end) {
```
gc00, line 106: Please add:
```
    char *msg[] = "*** Lower half called munmap, and it's not from the reserved memRange.";
    write(2, msg, sizeof(msg));
```
Originally, we were sloppy, in order to work quickly.  But now, we want to document any unusual behavior so that we can easily review the logic.

gc00 (approved):
I'm approving this in advance.  But please add the requested lines:
1.    Please see the comment about adding a write statement to document the behavior.
2.    Also, the second commit changes the dmtcp submodule.  But this is hidden inside a commit that has a completely different purpose.  Could you please change the dmtcp submodule in a _third_ commit, with a commit message explaining the reason?

Also, I had to restore `highMemStartStr`.  The original commit had removed it, and it didn't compile.  I've added a fixup commit, and rebased it back into the first commit.  If you work from a local copy, please fetch the latest version of this PR first.

PR #388, opened 2023-11-13
Fixing bugs related to the lower-half memory layout

xuyao0127:
This PR include multiple commits from the dev/gdc/simplifiedCopyBits testing branch. These commits fixes bugs on restarts that are related to lower-half memory layout issues.


gc00 (approved):
LGTM

PR #390, opened 2023-11-13
revert dmtcp coordinator

gc00:
dmtcp_coordinator:  tried to add --stale-timeout, but it didn't work well. See the comments in dmtcp_coordinator.cpp in PR #1090, where it says:
  * Also add FIXME comment about cleaner future logic


xuyao0127 (approved):
LGTM

PR #392, opened 2023-11-13
Add page after heap and change sbrk wrapper to return -1 for mtcp_restart, lh_proxy, upper half; after libc_start_main

gc00:
Without this fix, the restart or upper half can crash with libc:sysmalloc on stack when sysmalloc tries to call libc:morecore, which tries to call sbrk to extend the main (original) heap.


xuyao0127 (approved):
LGTM

PR #393, opened 2023-11-22
Add FI_MR_CACHE_MONITOR=memhooks in mana_restart

xuyao0127:
We also need the `FI_MR_CACHE_MONITOR=memhooks` environment variable in the mana_restart script for Perlmutter.  This was done for Perlmutter in the mana_launch script in commit 823aef0.


gc00 (approved):
LGTM.

PR #396, opened 2023-12-15
Conditional flag removal

aeblyve:
Conditionally remove `MAP_FIXED_NOREPLACE`

Without this, MANA does not compile on NEU Discovery.


gc00 (approved):
LGTM

PR #397, opened 2023-12-21
Reduce the MAX_DRAIN_ROUNDS for collectives to 200

xuyao0127:
Reduce the threshold used in PR#395 to reduce the freezing time.


gc00 (approved):
LGTM

PR #405, opened 2024-09-12
Remove --with-mana-helper-dir configure option

xuyao0127:
Fix a configuring error. Need for 1.0.1.

### DIFF
--- a/configure
+++ b/configure
@@ -5248,7 +5248,6 @@ ac_link='$CC -o conftest$ac_exeext $CFLAGS $CPPFLAGS $LDFLAGS conftest.$ac_ext $
 ac_compiler_gnu=$ac_cv_c_compiler_gnu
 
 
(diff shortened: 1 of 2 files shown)


xuyao0127:
> I'm seeing a branch `1.0.1_fix`. But I assume that the version will remain as 1.0.0. (We don't need to do a new release.)
> 
> I suspect that this PR won't change the behavior. Previously, it would assgn to `ac_configure_args`, and then assign again to `ac_configure_args`. If you delete the first assign, it should not change the behavior.
> 
> Anyway, LGTM.

When we were using the static lower half, DMTCP needed the --wtih-mana-helper-dir option to find the restart plugin directory. With the dynamic lower half, we no longer need this option. Also, DMTCP no longer supports it. So when Grace ran ./configure, the configuration failed because DMTCP couldn't recognize this option. That's why I need to remove them from MANA's configure script and want a new minor release.

PR #406, opened 2024-09-19
Fix for Discovery machine (CentOS 7)

xuyao0127:
This PR fixes multiple issues on the Discovery cluster.
1. Provide an alternative function on machines that don't support MAP_FIXED_NOREPLACE.
2. Remove unused test cases.
3. Fix how some Python script finds python3.
4. Remove unused wrapper functions.
5. Fix building errors.
6. Fix a bug in the kernel loader (wrong stack size for copying).
7. Add a feature for mana_launch.py and mana_restart.py scripts. The added --gdb option allows users to launch their program and MANA under gdb.

### DIFF
--- a/bin/mana_launch.py
+++ b/bin/mana_launch.py
@@ -15,6 +15,7 @@ dmtcp_options = ["-i", "--interval", "-h", "--coord-host",
                  "--ckpt-signal", "--with-plugin",
                  "--tmpdir", "--coord-logfile"]
 verbose = False
+gdb = False
 mana_root_path = os.path.dirname(os.path.realpath(__file__)) + "/../"
 
 # if no arguments provided, print the help message
@@ -42,6 +43,9 @@ if not target_app[0]:
   sys.exit(1)
 
 # Special arguments
+if "--gdb" in dmtcp_flags:
+  gdb = True
+  dmtcp_flags.remove("--gdb")
 if "--help" in dmtcp_flags:
   print(help_msg)
   sys.exit(1)
@@ -108,11 +112,15 @@ if coordinator_found != 0:
 # Build the command line to be executed
 if not verbose:
   dmtcp_flags.insert(0, "-q -q")
-cmd_line = mana_root_path + "bin/dmtcp_launch --mpi" + host_flag + port_flag + \
-           "--no-gzip --join-coordinator --disable-dl-plugin" + \
-           " --with-plugin " + mana_root_path + "lib/dmtcp/libmana.so " + \
-           " ".join(dmtcp_flags) + " " + mana_root_path + "bin/kernel-loader " + \
-           " " .join(target_app)
+if gdb:
+  cmd_line = shutil.which("gdb") + " --args "
+else:
+  cmd_line = ""
+cmd_line += mana_root_path + "bin/dmtcp_launch --mpi" + host_flag + port_flag + \
+            "--no-gzip --join-coordinator --disable-dl-plugin" + \
+            " --with-plugin " + mana_root_path + "lib/dmtcp/libmana.so " + \
+            " ".join(dmtcp_flags) + " " + mana_root_path + "bin/kernel-loader " + \
+            " " .join(target_app)
 if verbose:
   print(cmd_line)
 
--- a/mpi-proxy-split/lower-half/Makefile
+++ b/mpi-proxy-split/lower-half/Makefile
@@ -41,7 +41,7 @@ default: kernel-loader
 .c.o:
 	${MPICXX} ${CFLAGS} -c -o $@ $<
 
-kernel-loader: get_symbol_offset.o copy_stack.o patch_trampoline.o lh_func_ptr.o mem_wrapper.o switch_context.o kernel_loader.o ${DMTCP_DEPS}
+kernel-loader: get_symbol_offset.o copy_stack.o patch_trampoline.o lh_func_ptr.o mem_wrapper.o switch_context.o kernel_loader.o mmap-fixed-noreplace.o ${DMTCP_DEPS}
 	${MPICXX} -g3 ${TEXTSEG_ADDR_FLAG} ${CXXFLAGS} -o $@ $^ -lpthread -ldl
 
 install: kernel-loader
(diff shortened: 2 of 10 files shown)


mpi-proxy-split/lower-half/kernel_loader.cpp
```diff
@@ -119,22 +116,6 @@ int main(int argc, char *argv[], char *envp[]) {
   memset(lh_info, 0, sizeof(LowerHalfInfo_t));
   lh_info->fsaddr = (void*)fsaddr;
```
gc00: I forget the details here.  I remmeber that both DMTCP and MANA were creating a memory segment just above the end of data (above `sbrk(0)`).  This was to make sure that libc.so doesn't try to extend thee alloc arena, so as to simplify the logic.

I'm assuming that DMTCP does it now, and MANA doesn't need to now?  Is that correct?  Could you reply here on github, so that we have a permanent record of our thinking?

mpi-proxy-split/lower-half/mem_wrapper.cpp
```diff
@@ -329,58 +329,3 @@ void updateMmaps(void *addr, size_t length) {
     }
   }
 }
```
gc00, line 331: QUESTION:  I assume that we don't need this now because either DMTCP or MANA adds a page just beyond `sbrk(0)`, so that extending the heap is no longer possible.  Is that true?

If that's true, could you add a short comment here (in `mem_wrapper.cpp`) saying that:
> // We don't need a wrapper for sbrk() because [DMTCP? MANA?] has created a page above here to make sbrk(0) fail in libc/application.

mpi-proxy-split/lower-half/Makefile
```diff
@@ -42,7 +42,7 @@ default: kernel-loader
 	${MPICXX} ${CFLAGS} -c -o $@ $<
 
 kernel-loader: get_symbol_offset.o copy_stack.o patch_trampoline.o lh_func_ptr.o mem_wrapper.o switch_context.o kernel_loader.o mmap-fixed-noreplace.o ${DMTCP_DEPS}
```
gc00, line 44: This is fine.  Just a comment for your own information.  Normally, gcc (or clang) will automatically link with libatomic.so.  But for some MPICXX it's hard to know if they cover it.  So, this is good as a safety measure.

mpi-proxy-split/lower-half/copy_stack.c
```diff
@@ -84,7 +84,7 @@ char *deepCopyStack(int argc, char **argv, char *argc_ptr, char *argv_ptr,
                         32 /* max.  padding is 32 for x86_64 */ +
                         auxv_size + env_ptr_size + argv_ptr_size +
                         (auxv_size + env_ptr_size) % 16 /* 16-byte aligned */ +
```
gc00, line 90: Aha!  I fell into a trap that you prepared for me.  :-)

You have a commit with msg:
>Fix bug when kernel loader copy stack
And you have a commit with msg:
> Fix bug that kernel loader uses wrong argc addr

Can you please combine the two commits, since it's two fixes for a `single` piece of code.

----
**This is my old comment from the first commit.  Please ignore it:**


As we discussed, argv_ptr - argc_ptr should have been small.  It's large because `_start()` copied argc to a second address, which is passed to `main()`.  But the second address doesn't exist when we load (i.e., kernel-loader).

So, let's at least add a comment to that effect.  But even better, we can copy some of kernel-loader-v12.tgz (copy-stack.c) here:
```
  // The fnc _start() may copy 'argc', but this is its original address.
  char *argc_ptr_original = argv_ptr - sizeof(char *);
  // This shows that for current stack (not the copied one), _start copied argc.
  assert(*(long int *)argc_ptr_original == *(long int *)argc_ptr);
  ...
              (argv_ptr - argc_ptr_original); 
```

mpi-proxy-split/uh_wrappers.cpp
```diff
@@ -96,7 +96,15 @@ int munmap(void *addr, size_t length) {
 static void readLhInfoAddr() {
   char filename[100] = "./lh_info_";
   gethostname(filename + strlen(filename), 100 - strlen(filename));
-  snprintf(filename + strlen(filename), 100 - strlen(filename), "_%d", dmtcp_get_real_pid());
+  filename[strlen(filename)] = '_';
+  // convert real pid to char* without calling snprintf
```
gc00, line 100: Probably capitalize, above:  `convert` -> `Convert`.
And add a comment:
```
    // During process startup, avoid calling a libc fnc that is a DMTCP wrapper
```

bin/mana_launch.py
```diff
@@ -42,6 +43,9 @@
   sys.exit(1)
 
 # Special arguments
+if "--gdb" in dmtcp_flags:
+  gdb = True
+  dmtcp_flags.remove("--gdb")
```
gc00, line 48: This is minor, but for style, show about move the `gdb = False` down here with:
```
else:
  gdb = False
```

Then the code is local.  I hate having to search multiple places in the code to understand what it does.

mpi-proxy-split/lower-half/Makefile
```diff
@@ -41,7 +41,7 @@ default: kernel-loader
 .c.o:
```
gc00, line 41: Please fix the spelling of this commit message:
```
Add mmap_fixed_noreplace for the kernel laoder
->
Add mmap_fixed_noreplace for the kernel loader
```

gc00 (approved):
I'm approving this in advance.  But please address my various (small) comments.

Thanks very much for this important commit!

mpi-proxy-split/lower-half/Makefile
xuyao0127: It's automatically linked on Perlmutter. However, on Discovery, the compiler complains about undefined functions in the library. So I add `-latomic` here for safety.

bin/mana_launch.py
xuyao0127: The `gdb` variable is initialized to `False` at the beginning of the script. So we don't need another else clause.

mpi-proxy-split/lower-half/kernel_loader.cpp
xuyao0127: For now, DMTCP, MANA's upper half, and the kernel loader all have this functionality. I think it's safe to remove the corresponding parts in MANA's upper half and the kernel loader. But I haven't tested. We can do this later.

PR #409, opened 2024-11-20
Update kernel-loader to v12a

xuyao0127:


xuyao0127:
1. In the lower-half directory, there's a VERSION file indicating the version of the kernel-loader.
2. I just changed all files in the lower-half directory to use "-" instead of "_". But there are many other files in MANA that use "_" in the file name. We can update them later.

PR #412, opened 2024-12-23
Fixed restart bug related to upper-half stack segments

xuyao0127:
This PR fixed several bugs related to mismarking upper-half memory segments. This PR also upgraded the DMTCP version to 3.1.2.


gc00 (approved):
LGTM.

PR #413, opened 2024-12-23
New p2p

xuyao0127:
Improved MPI point-to-point communication wrapper and draining algorithm.

This PR is marked as a draft because the performance is unstable, especially with a large message size. Also because this PR is based on [PR#412 ](https://github.com/mpickpt/mana/pull/412) which is still under review.

### DIFF
--- a/mpi-proxy-split/p2p_drain_send_recv.cpp
+++ b/mpi-proxy-split/p2p_drain_send_recv.cpp
@@ -24,11 +24,16 @@
 #include <map>
 #include <algorithm>
 #include <vector>
+#include "kvdb.h"
 #include "jassert.h"
 #include "p2p_drain_send_recv.h"
 #include "p2p_log_replay.h"
 #include "mpi_nextfunc.h"
 #include "virtual_id.h"
+
+using namespace dmtcp;
+using dmtcp::kvdb::KVDBRequest;
+using dmtcp::kvdb::KVDBResponse;
 
 extern int MPI_Alltoall_internal(const void *sendbuf, int sendcount,
                                  MPI_Datatype sendtype, void *recvbuf,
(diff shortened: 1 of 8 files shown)


gc00:
@xuyao0127 , This PR has been around for a while.  Do we still need it, or should we close it, in favor of merging your development branch at a future time?

gc00:
@xuyao0127, If this PR is now current, could you rebase for the current MANA?

xuyao0127:
> @xuyao0127, If this PR is now current, could you rebase for the current MANA?

@gc00 I just rebased this PR. Can you review it?

mpi-proxy-split/p2p_drain_send_recv.cpp
```diff
@@ -257,6 +247,16 @@ drainSendRecv()
...
+    // If pending MPI_Irecv or MPI_Isend, use MPI_Test to try to complete it.
+    completePendingP2pRequests();
+    // If MPI_Irecv not posted but msg was sent, use MPI_Iprobe to drain msg.
+    drainRemainingP2pMsgs();
+    // Update global recv coutner.
+    registerLocalSendsAndRecvs();
+  }
 }
 
 // FIXME: isBufferedPacket and consumeBufferedPacket both search
```
gc00, line 262: I know that the name `isBufferedPacket()` was here before the PR.  But let's change the name now, to make it more redable.
Maybe we can think of a better name.  A packet is a network packet that is only part of a message.  I propose `MsgBuffer` instead of `BufferedPacket`.
Since we are inspecting messages in g_message_queue, we should probably say `Message` or `Msg` instead of `Packet`.

How about using a name like:
> `existsMatchingMsgBuffer()`
And then we should also change `consumeBufferedPacket` to:
> `consumeMatchingMsgBuffer()`

[ I write consume _Matching_ msg since that function also checks for a match. ]

mpi-proxy-split/test/Makefile
```diff
@@ -42,6 +42,7 @@ FILES=mpi_hello_world \
       Irecv_test Alloc_mem sendrecv_replace_test \
       Type_hvector_test Type_vector_test \
       ${FILES_FORTRAN} \
+      send_recv \
```
gc00, line 45: It's minor, but why not keep `send_recv` and `send_recv_loop` on the same line.  Let's collapse this line into the next line.

gc00 (approved):
I'm approving in advance.  But please read the two inline comments about small changes.

gc00:
I'm just checking back.  I think we still need the minor changes that I proposed (e.g., MsgBuffer instead of BufferPacket, etc.).  Thanks.

PR #414, opened 2025-01-15
Fix deepcopy discovery

rajatpratapbisht:
# obj: run mana on discovery cluster
1.  `Added Datatype for mana_handle union memeber for MPI_FILE`: added datatype signature  for vanilla MPI lib..
2. `Commented out Fortran f90 files from test Makefile`: fortran version incompatible on discovery cluster. All systems will not not have Fortran test files..
3. `fixed 16-bit alignment error on Discovery`: STACK_RLIMT.rlimit_cur is set to RLIMIT_INFIITY by default on discovery. This commit tests for this edge case and uses 32MB as rlimit_cur (soft limit) value. Without this commit test for 16-bit alignment fails for deepCopyStack()..

### DIFF
--- a/mpi-proxy-split/lower-half/kernel-loader.cpp
+++ b/mpi-proxy-split/lower-half/kernel-loader.cpp
@@ -102,8 +102,6 @@ void set_addr_no_randomize(char *argv[]) {
 }
 
 int main(int argc, char *argv[], char *envp[]) {
-  
-
   set_addr_no_randomize(argv);
   int i;
   int restore_mode = 0;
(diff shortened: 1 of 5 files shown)


xuyao0127 (commented):
There are some empty lines in this commit. Can you remove them?

mpi-proxy-split/lower-half/kernel-loader.cpp
```diff
@@ -102,6 +102,8 @@ void set_addr_no_randomize(char *argv[]) {
 }
 
 int main(int argc, char *argv[], char *envp[]) {
+  
+
```
xuyao0127, line 106: There are some empty lines in this commit. Can you remove them?

rajatpratapbisht: Fixed it in the latest commit

xuyao0127 (approved):
LGTM

PR #415, opened 2025-01-15
Fix mana restart

rajatpratapbisht:
1.` Fixed --restartdir <dir> flag option in mana_restart.py` :  Fixed mana_restart.py code to be able to load targeted restart dir path using --restartdir flag.
2. `FIXED inline --gdb flag crashing bug in mana_restart.py` : Fixed freezing error caused when inline call to gdb using --gdb flag is used.

### DIFF
--- a/bin/mana_restart.py
+++ b/bin/mana_restart.py
@@ -11,7 +11,7 @@ help_msg = '''
 
 # dmtcp options that requires a value
 verbose = False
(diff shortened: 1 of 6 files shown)


xuyao0127:
LGTM.

xuyao0127 (approved):
LGTM

PR #416, opened 2025-01-18
New mmap wrapper

xuyao0127:
This new upper-half mmap wrapper allocates memories starting at 3GB. This region is far from the memory segments of the lower-half libraries. By doing so, MANA can avoid potential memory conflicts during restart.


mpi-proxy-split/lower-half/mem-wrapper.cpp
```diff
@@ -47,35 +47,28 @@ using namespace std;
 #define _real_mmap mmap
 #define _real_munmap munmap
 
+#define UH_BASE_ADDR ((void *)0xC0000000) // 3GB
```
gc00, line 50: As discussed together, maybe this should first look for an environment variable, like `MANA_UH_BASE_ADDR`.  Then it's easy for a user to change this.  But I'll approve this in advance.

gc00 (approved):
I'm approving this in advance.  Please see the inline comment.  Thanks!

mpi-proxy-split/lower-half/mem-wrapper.cpp
```diff
@@ -130,6 +123,21 @@ static void* __mmap_wrapper(void *addr, size_t length, int prot,
                            int flags, int fd, off_t offset) {
   static void *libc_base_addr = NULL;
   void *ret = MAP_FAILED;
+  if (flags & MAP_FIXED) {
+    if (addr > curr_uh_free_addr) {
+      get_next_addr(length);
+    }
+    DLOG(INFO, "User calls mmap with MAP_FIXED\n");
+  } else if (flags & MAP_FIXED_NOREPLACE) {
```
xuyao0127, line 131: Test if MAP_FIXED_NOREPLACE is defined before checking the flag.

PR #418, opened 2025-02-10
CoMD debug

rajatpratapbisht:
**`This PR includes the following commits`**:
* `New test for MPI_Double_Int datatype with MPI_Allreduce`: MANA fails to resolve MPI_Double_Int for a call to  MPI_Allreduce  CoMD software. This commit adds a small and faster toy program to MANA's test directory for quickly filtering similar failures in future.
* `Fixed upper_to_lower_constants lookup for special(negative) Datatypes`: Special Datatypes are defined in MPI library for 'struct' type datatypes, defined as negative constant values. This commit aims to maintain  integrity of these special(negative) datatype when passed in function calls.

### DIFF
--- a/mpi-proxy-split/test/MPI_Double_Int_test.c
+++ b/mpi-proxy-split/test/MPI_Double_Int_test.c
@@ -1,0 +1,30 @@
+#include <stdio.h>
+#include <mpi.h>
+
+int main(int argc, char** argv) {
+  MPI_Init(&argc, &argv);
+
+  int rank, size;
+  MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+  MPI_Comm_size(MPI_COMM_WORLD, &size);
+
+  struct {double value; int rank;} local_data, global_data;
+
+  // assign a value (eg: some computational result)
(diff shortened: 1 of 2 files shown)


mpi-proxy-split/test/MPI_Double_Int_test.c
```diff
@@ -0,0 +1,30 @@
+#include <stdio.h>
+#include <mpi.h>
+
```
gc00, line 3: Could you add a short comment here:  MPICH treats this as a negative 32-bit integer, when it should be unsigned int.  So, MANA used to sign extend this to 64 bits with the wrong value.  This tests for a possible regression.

gc00 (approved):
I'm approving this in advance.  Please note my two comments and fix them.

Also, after pushing this in, please let @xuyao0127 know, so that he can add this commit to his development branch.

mpi-proxy-split/test/MPI_Double_Int_test.c
rajatpratapbisht: done

PR #421, opened 2025-02-25
unsynchronized MPI_Finalized  

rajatpratapbisht:
This PR is based on PR#374 created by Leonid Belyaev (@aeblyve). 

**Unsynchronized MPI_Finalize**  call to MANA wrappers causes MPI application to prematurely terminate i.e., it `aborts` any slacking process still computing and the result is lost.
 
This PR fixes this bug with the following commits:
* `Added test for unsynchronized MPI_Finalize call`: Added a test in MANA test directory to detect future regression. A
* `Fix MPI_Finalize wrapper to correctly call MPI_Finalize`:  Added code in MANA wrapper for MPI_Finalize to call MPI_Finalize from lower half instead of returning MPI_SUCESS. This allows all the running processes to finish safely without premature termination.
* `Removed stale comment from MPI_Finalize wrapper`: Stale comment regrading Cray-MPI removed form MPI_Finalize wrapper, since it is no longer relevant. 
* `Corrected indentation in Makefile for test dir`: Fixed layout in Makefile for test directory.

### DIFF
--- a/mpi-proxy-split/mpi-wrappers/mpi_wrappers.cpp
+++ b/mpi-proxy-split/mpi-wrappers/mpi_wrappers.cpp
@@ -104,15 +104,6 @@ USER_DEFINED_WRAPPER(double, Wtime, (void))
 
 USER_DEFINED_WRAPPER(int, Finalize, (void))
 {
-  // FIXME: With Cray MPI, for some reason, MPI_Finalize hangs when
-  // running on multiple nodes. The MPI Finalize call tries to pthread_cancel
-  // a polling thread in the lower half and then blocks on join. But,
-  // for some reason, the polling thread never comes out of an ioctl call
-  // and remains blocked.
-  //
-  // The workaround here is to simply return success to the caller without
-  // calling into the real Finalize function in the lower half. This way
-  // the application can proceed to exit without getting blocked forever.
   int retval;
   DMTCP_PLUGIN_DISABLE_CKPT();
   JUMP_TO_LOWER_HALF(lh_info->fsaddr);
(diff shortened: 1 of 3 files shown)


mpi-proxy-split/mpi-wrappers/mpi_wrappers.cpp
```diff
@@ -113,7 +113,13 @@ USER_DEFINED_WRAPPER(int, Finalize, (void))
   // The workaround here is to simply return success to the caller without
```
gc00, line 113: I think these comments (starting at line 107, above) are obsolete, with your fix.  Please change the comments.

gc00 (commented):
Please change the comments to update them, and let me know.  I'll then give it a quick new review.

gc00:
@rajatpratapbisht , Please note that I'm asking about the comments starting at line 107 of mpi_wrappers.cpp.   Is that comment now obsolete?  If so, please delete it or modify it.  You can answer in the "conversation" for the inline comment.

mpi-proxy-split/mpi-wrappers/mpi_wrappers.cpp
rajatpratapbisht: Thank you Gene for this catch. I have updated it now.

mpi-proxy-split/test/Makefile
```diff
@@ -43,10 +43,11 @@ FILES=mpi_hello_world \
       Type_hvector_test Type_vector_test \
       ${FILES_FORTRAN} \
       send_recv_loop large_async_p2p \
```
gc00, line 45: In the four modified lines below, you seem to have changed the formatting style from the original author.
The original author chose to indent by 4 spaces.

When you change trivial formatting, be careful if there is no important reason.  Otherwise, when anyone in the future does a diff, they will have to again read your decision to change the formatting style from the original.

gc00 (approved):
I'm approving this in advance.  But please see my minor inline comment, and consider fixing that.

mpi-proxy-split/test/Makefile
rajatpratapbisht: Thank you for pointing this out. :)
Original author used 6 'single-spaces' instead of 'tab' space. Using 'tab' instead of 'single-space' changes the file layout but showed no difference in my VIM. LOL!

PR #422, opened 2025-03-04
UCX library constructor fix (open source)

rajatpratapbisht:
Network libraries(for example intel's implementation of ucx library used for InfiniBand) that use constructor functions to run before ``main`` of Upper-Half MPI applications, caused crash under MANA. 
This PR aims to fix that by providing a solution for open-source upper-half MPI-applications by the help of following commits:
* `Updated correct compiler flag in test dir Makefile`: Compiler for MPI applications  linked to **_mpistub_** library now set to compile with  **_gcc_** instead of **_mpicc_**.  
* `Update LD_LIBRARY_PATH with libmpistub for UH`: Added code to update **LD_LIBRARY_PATH** for upper-half by prepending path to file **libmpistub.so**.
* `Added custom mpicc_mana compiler in bin dir`:  For open-source MPI applications, this commit provides a custom compiler.  This compiler is to be used on systems that provide **_mpicc_** implementation dependent on libraries with constructor functions. This custom compiler `mpicc_mana` will compile source code with libmpistub library, instead of system MPI libraries, making it suitable to run under MANA checkpointing and restart software.
* `Fixed mpicc_mana to only support dynamic libraries`:  mpicc_mana only supports libmpistub.so, which is a dynamic library. This commit makes mpicc_mana to exit after printing an error message for static linking requests.

### DIFF
--- a/bin/mpicc_mana
+++ b/bin/mpicc_mana
@@ -52,6 +52,8 @@ for arg in "$@" ; do
       interlib_deps=no
       static_mpi=yes
       addarg=no
+      echo "Does not support Static Library"
+      exit 0
     ;;
     -show)
       addarg=no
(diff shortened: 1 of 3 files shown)


xuyao0127 (approved):
LGTM

PR #423, opened 2025-03-07
Ucx constructor fix for closed-source MPI applications

rajatpratapbisht:
MPI applications compiled with dynamic libraries, containing constructors to be run before main() creates a problem with running under MANA checkpointing and restarting software (example: 'intel' and 'ucx'). 
This PR proposes a fix by creating Shadow library with same symbolic link names but linked to a dummy library instead, enough to avoid linker-loader's error. 
The following commits aim to achieve this patch:
* `Added new Python script for shadow library creation`: A new python script is added to `bin` directory, which uses dependency tree, and filters out **intel** and **ucx** libraries. These filtered out library names are then used to create a shadow library with dummy symbolic links to a dummy library instead of correct libraries.
* `Added --use-shadowlibs feature support to mana_launch.py`: when launching MPI-applicaiton under MANA with mana_launch.py, this commits adds support for **`--use-shadowlibs`** flag. This flag allows `mpi_launch.py` to create a shadow library for the current MPI-executable. 
* `Added support for prepending shadow lib path to LD_LIBRARY_PATH`: Lower-half application now checks for the presence of shadow library and prepends it to its LD_LIBRARY_PATH env variable, before deep-copying upper-half user application.

### DIFF
--- a/bin/mana_launch.py
+++ b/bin/mana_launch.py
@@ -2,10 +2,11 @@
 import sys
 import os
 import shutil
+import subprocess
 
 help_msg = '''
   USAGE: [srun] mana_launch [--verbose] [--timing] [DMTCP_OPTIONS ...]
-                                         [--ckptdir DIR] MPI_EXECUTABLE [...]
+                            [--use-shadowlibs] [--ckptdir DIR] MPI_EXECUTABLE [...]
           For DMTCP options, do: dmtcp_launch --help
 '''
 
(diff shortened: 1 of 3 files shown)


gc00 (changes requested):
Please update this PR, and I'll quickly approve it.  Thanks.

gc00 (approved):
LGTM.

PR #424, opened 2025-03-07
Fsgsbase fix

xuyao0127:
In the past, both the upper and lower half needed to check if the fsgsbase instructions are supported. And the upper half has a bug setting FsGsBaseEnabled.


gc00 (approved):
LGTM

PR #425, opened 2025-03-09
Fix compiler warning about MPI_Fnc_strings unused

gc00:
Fix a compiler warning about MPI_Fnc_strings unused.
If it's used in a .cpp file, then we set:
> `#define USES_MPI_Fnc_strings`

before
> `#include "lower-half-api.h"`


xuyao0127 (approved):
LGTM

PR #429, opened 2025-03-11
Mana mpi handle fix

rajatpratapbisht:
following 4 commits:
* `Added mana_handle type casting in wrappers`:  typecasting to mana_handle is useful for earlier versions of gcc.
* `Replaced 'mana_handle' with 'mana_mpi_handle'`: to avoid confusion, we want to specify that the handles used are for mpi under mana.
* `Added support for parsing inline CPP comments`: needed support for inline comments(`//`) and '#if 0' block
* `Removed usage of functions removed in MPI-3.0`: deprecated functions in MPI-2.0 are now removed completely in MPI-3.0. Therefore we have removed their support as well as updated code where they were used.   We retain some MANA wrappers for the removed functions, but they are markde `#if 0 ... #endif`, and can be removed later if there are no problems.


gc00 (approved):
Minor suggestion.
But otherwise, LGTM.  Approving now.

PR #431, opened 2025-03-12
Removed obsolete function support in lowerhalf

rajatpratapbisht:
Lower half API also consisted of obsolete functions which have been removed from MPI-3.0 standard.


mpi-proxy-split/lower-half/lower-half-api.h
```diff
@@ -167,7 +167,7 @@ typedef struct _LowerHalfInfo
 } LowerHalfInfo_t;
 
 extern LowerHalfInfo_t *lh_info;  
-
+#if 0
```
gc00, line 170: Instead of `#if 0' for all these `MACRO()` commands, please just delete the specific `MACRO()` calls that are now obsolete.  This will be a much smaller change.  And in the unlikely event that we wish to bring back an obsolete function, we will know how to insert again a `MACRO()` call, based on compiler warnings and errors,

gc00 (approved):
But please note my inline comment:  Instead of making the large change to the list of `MACRO()` calls, please just keep the original list, and delete the `MACRO()` calls that are now obsolete, and have been removed.

Also, one minor point.  It's not important, but in your commit message, I would have said:
> Commented out obsolete function support in lowerhalf

That still fits in the 52 char's for the commit msg, and it's more precise.  :-)

mpi-proxy-split/lower-half/lower-half-api.h
rajatpratapbisht: thank you Gene.
I have updated my changes as per your suggestions. :)

PR #432, opened 2025-03-13
Fix compiler warnings

gc00:
We had a huge number of compiler warnings in lower-half dir (and some elsewhere).
I got bored with fixing the any warnings, but this fixes some of the worst of them.  This would be embarrassing for our release.

(See the first commit msg for an analysis of the many warnings.)

This is ready for review now.

### DIFF
--- a/mpi-proxy-split/lower-half/lower-half.cpp
+++ b/mpi-proxy-split/lower-half/lower-half.cpp
@@ -1,20 +1,19 @@
-#include <asm/prctl.h>
-#include <sys/syscall.h>
-#include <unistd.h>
+#include <assert.h>
 #include <dlfcn.h>
+#include <elf.h>
+#include <errno.h>
+#include <fcntl.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <unistd.h>
 #include <string.h>
-#include <sys/types.h>
-#include <sys/stat.h>
 #include <sys/personality.h>
-#include <fcntl.h>
+#include <asm/prctl.h>
+#include <sys/stat.h>
+#include <sys/syscall.h>
+#include <sys/types.h>
 #include <sys/mman.h>
 #include <sys/resource.h>
-#include <assert.h>
-#include <errno.h>
-#include <elf.h>
 
 #include "mtcp_header.h"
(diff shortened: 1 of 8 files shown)


mpi-proxy-split/lower-half/lower-half.cpp
```diff
@@ -833,14 +831,16 @@ static int restoreMemoryArea(int fd, MtcpHeader *ckptHdr)
   ProcMapsArea area;
 
   int rc = read(fd, &area, sizeof area);
+  assert(rc == sizeof area);
+  JASSERT(mprotect(area.addr, area.size, area.prot) == 0);
```
xuyao0127, line 835: At this point, MANA is reading the "area" structure from the checkpoint image file. The area is not in the memory yet (mapped at line 939 in this version). So calling protect on `area.addr `will fail.

mpi-proxy-split/lower-half/lower-half.cpp
```diff
@@ -24,6 +23,8 @@
 #include "dmtcp.h"
 #include "dmtcprestartinternal.h"
 #include "procmapsarea.h"
+// FIXME:  lower-half should be a standalone program, except where it uses
```
xuyao0127, line 26: MANA doesn't use mtcp_restart. Instead, MANA uses functions in the `dmtcprestartinternal.h/cpp` file to restore the upper half. The problem was that MANA couldn't integrate with `dmtcp_restart` which uses `execvp`.  So Kapil provided APIs using the `dmtcprestartinternal.h` file. MANA can use these APIs to work as a "`dmtcp_resetart`".

The next step was integrating `mtcp_restart`. Compared to loading the mtcp_restart to the upper half and jumping to the entry point of mtcp_restart, it was easier to copy core codes from `mtcp_restart` to the lower-half program and restore memories in the upper-half directly. 

We can think about how to simplify the lower-half program. I agree it's bloated now. But we can only remove these dmtcp header files after the lower-half program at least no longer works as a "`dmtcp_restart`"

xuyao0127 (approved):
I'm approving in advance. Need to fix the bug in `lower-half.cpp`

PR #434, opened 2025-03-14
Simplified wrappers

xuyao0127:
Simplifies the wrapper structure in MANA. Fortran wrapper files and stub library files are autogenerated, so they are not touched.

### DIFF
--- a/mpi-proxy-split/mpi-wrappers/mpi_cart_wrappers.cpp
+++ b/mpi-proxy-split/mpi-wrappers/mpi_cart_wrappers.cpp
@@ -41,7 +41,8 @@ using namespace dmtcp_mpi;
 
 extern "C" {
 
-int MPI_Cart_coords(MPI_Comm comm, int rank, int maxdims, int *coords)
+#pragma weak MPI_Cart_coords = PMPI_Cart_coords
+int PMPI_Cart_coords(MPI_Comm comm, int rank, int maxdims, int *coords)
 {
   int retval;
   DMTCP_PLUGIN_DISABLE_CKPT();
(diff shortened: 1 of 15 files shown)


xuyao0127:
Commit 9748aab196555cc97f84b09d07cd3a1001ed0018 is also included in PR https://github.com/mpickpt/mana/pull/433. After merging that PR, we can rebase against the main branch so that this PR contains only one.

gc00 (commented):
QUESTION:  Were the PMPI macros implementing the PMPI family (profiling MPI)?  If so, do we need to keep the PMPI macros?

Also, I think for the first commit (type conversions), it probably duplicates my PR #432, but in a different way.  The philosophy of PR #432 is that we use 'char *' if we might need to do arithmetic on the types.  We use 'void *' if we just want to pass the pointer to other functions, but we weill never do arithmetic on those pointers.

xuyao0127:
@gc00 According to the MPI 4.1 Section 15.2:

> To meet the requirements for the MPI profiling interface, an implementation of the MPI functions must provide a mechanism through which all of the MPI defined functions, except those
> allowed as macros (See Section 2.6.4), may be accessed with a name shift. This
> requires, in C and Fortran, an alternate entry point name, with the prefix PMPI_ for
> each MPI function in each provided language binding and language support method.

Further in Section 15.2.5, the standard suggested that the MPI implementation should define PMPI_ names as weak symbols of MPI_ names.

For normal applications, not supporting PMPI layer is fine. If we also want to support profiling tools inserted between user's application and MANA, we should also define PMPI symbols.

gc00 (approved):
@xuyao0127 , I'll trade you.  I'm approving your PR #434.  Could you review my PR #432 ?  Thanks.

xuyao0127:
@gc00 Can you have a final look at this?

gc00:
Thanks!  I looked at the PMPI issue, and it looks good to me.  Let's push it in.

PR #435, opened 2025-03-14
Minor fixes for shadow library support

rajatpratapbisht:
This PR improves upon previous version of implementing usage of shadow-libs with the following 2 commits:
* `Fix: shadow-lib deletion error (race-condition)`: While deleting shadow libraries with mana_launch.py, due to multiple processes trying to remove same files, we were seeing run time errors and process exits prematurely. This race condition error is now resolved by adding a `try: ... catch e: ` to gracefully deal with them.
* `FIXUP: update LD_LIBRARY_PATH only at launch`: We now update the LD_LIBRARY_PATH only when we are in LAUNCH state.

### DIFF
--- a/mpi-proxy-split/lower-half/lower-half.cpp
+++ b/mpi-proxy-split/lower-half/lower-half.cpp
@@ -294,9 +294,6 @@ int main(int argc, char *argv[], char *envp[]) {
   lh_info->MANA_ERRORS_ARE_FATAL = MPI_ERRORS_ARE_FATAL;
   lh_info->MANA_ERRORS_RETURN = MPI_ERRORS_RETURN;
 
-  // Update LD_LIBRARY_PATH to contain
-  update_library_path(argv[0]);
-
   // Check arguments and setup arguments for the loader program (cmd)
   // if has "--restore", pass all arguments to mtcp_restart
   for (i = 1; i < argc; i++) {
@@ -418,6 +415,12 @@ int main(int argc, char *argv[], char *envp[]) {
     postRestart(0, 0);
     // The following line should not be reached.
     assert(0);
+  } else {
+    // In LAUNCH mode:
+    //  Prepend LD_LIBRARY_PATH env variable to
+    //    * '/PATH_TO_MANA/lib/dmtcp' for libmpistub.so (always)
+    //    * '/PATH_TO_MANA/lib/tmp'   for shadow libraries (only when launched with --use-shadowlibs flag)
+    update_library_path(argv[0]);
   }
 
   for (i = 1; i < argc; i++) {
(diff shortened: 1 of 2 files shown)


mpi-proxy-split/lower-half/lower-half.cpp
```diff
@@ -418,6 +415,12 @@ int main(int argc, char *argv[], char *envp[]) {
     postRestart(0, 0);
     // The following line should not be reached.
     assert(0);
+  } else {
```
xuyao0127, line 418: I don't think you need this else. Because if it's restart, this program will not execute beyond line 415. Actually all codes after this line are for launching.

rajatpratapbisht: I agree with you. But for programmers going through our code to understand the use of 'shadow libs' this adds readability to have it in 'else'. This will assert that LD_LIBRARY_PATH is only updated in launch while adding no overhead to our application.

gc00: If we're talking about readability, we've lost the battle already.  The "if" branch covers 105 lines  And after those 105 lines, we then have an "else" clause here.  And the beginnign of the function is 250 lines earlier (about 5 screenfuls).  So, it's almost impossible to read unless we load into an editor, and used editing commands for "got-to-beginning-of-fnc".)

@rajatpratapbisht , I vote for approving this PR right now, but only on one condition.  :-)   Will you promise to do a new PR to revise this code?  The usual idea is to create well-named helpr frunctions, and then to call the helper functions, so that the "if" and "else" are reasonably close to each other in the code.

If you're willing to do a follow-on PR, then you could add a comment in the "else" branch for now, explaining that in fact, there's nothign to do during restart.  Then we can push in this PR now.  Do you agree to this?

gc00: @rajatpratapbisht ,
    Great!  Then I'm approving this PR.
    But just a short comment.  A function name `conditional block to do restart` is a terrible function name.  What I'm really asking is for you to decompose the steps of those 100 or 250 lines of code into a high-level outline:  step-1-doX, step--2-do-Y, etc.  (But don't use "step-1" and "step-2" in your function names.  The function names should be something meaningful that allows us to understand the _algorithm, without having to wade through the details of each step of the algorithm.

Thanks.

gc00 (approved):
I'm approving, based on our bargain.

In a second PR, you will analyze the high level steps of the algorithm being used.  And then you'll give a meaningful function name for each step of the algorithm.  But you don't need a name include "step 1".  When you call the step-1 function first, and the step-2 functio second, it will already be obvious that these are Steps 1 and 2.

mpi-proxy-split/lower-half/lower-half.cpp
rajatpratapbisht: Yes Gene! :) 
Thank you.

PR #436, opened 2025-03-16
Lower half readability fix

rajatpratapbisht:
This PR extends PR#435 by providing helper functions in `lower-half.cpp` for improved readability.
Description for 2 commits is as follows:
* `Added helper function for RESTART mode execution`: Added helper functions to improve readability and reduce the code block for restart mode in main from 100+ lines to less than 10 lines, while maintaining readability of code. 
* `Added helper functions to parse command line args`: Added helper function to parse command line arguments for launch mode, and to remove repetition of code for printing help/usage message.

### DIFF
--- a/mpi-proxy-split/lower-half/lower-half.cpp
+++ b/mpi-proxy-split/lower-half/lower-half.cpp
@@ -73,16 +73,8 @@ static void patchAuxv(Elf64_auxv_t *av, unsigned long phnum,
 void *create_red_zone(char *stack_addr);
 void update_library_path(const char *argv0);
 int prepend_to_library_path(const char *new_path);
-
-// Restart Mode helper functions
-int parse_restore_flag(int *argc, char **argv);
-RestoreTarget* get_restore_target_info(DmtcpRestart &dmtcpRestart, int rank);
-void validate_checkpoint_header(RestoreTarget *t, MtcpHeader &ckpt_hdr);
-void reserve_memory_areas(RestoreTarget *t, off_t &ckpt_file_pos);
-void release_reserved_memory(RestoreTarget *t, off_t ckpt_file_pos);
-void restore_session_leadership(RestoreTarget *t);
-void restore_memory_data(RestoreTarget *t, MtcpHeader &ckpt_hdr);
-
+void print_usage_and_exit(char *prog_name);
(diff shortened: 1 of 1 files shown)


mpi-proxy-split/lower-half/lower-half.cpp
```diff
@@ -415,186 +332,172 @@ int main(int argc, char *argv[], char *envp[]) {
...
     //    * '/PATH_TO_MANA/lib/dmtcp' for libmpistub.so (always)
     //    * '/PATH_TO_MANA/lib/tmp'   for shadow libraries (only when launched with --use-shadowlibs flag)
     update_library_path(argv[0]);
-  }
-
-  for (i = 1; i < argc; i++) {
-    if (strcmp(argv[i], "-h") == 0) {
-      fprintf(stderr, "USAGE:  %s [-a load_address] <command_line>\n", argv[0]);
-      fprintf(stderr, "  (load_address is typically a multiple of 0x200000)\n");
+
```
gc00, line 350: Starting at the next line, and following until `exit(0)`, can we change this to a helper function, perhaps with the name:
> `load_upper_half_using_ld_so(int cmd_argc, char *cmd_argv[])`

I didn't test this function signature.  You mmight need additional arguments, also.

gc00 (commented):
Thanks, @rajatpratapbisht !  This definitely makes the code more readable.
I have one further suggestion for a helper function to make the code more readable.

@xuyao0127 , Could you do the final review of this code, since you had written a lot of it?  Please check that we are not introducing bugs in this code transfformation.

xuyao0127 (approved):
This looks good to me. Maybe we can extract "kernel-loader" codes in the launch case to a separate function. This may help integrate CRAC and refactor the restart process as Gene and I discussed. But we don't need it immediately.

rajatpratapbisht:
@xuyao0127 
I wanted to do the same for launch  with helper functions, but I wanted to have a discussion with you and Gene first.

mpi-proxy-split/lower-half/lower-half.cpp
rajatpratapbisht: I tried a few idea on launch section of code but I was getting some compilation errors. I will soon create a PR for 'helper functions in launch' after I do a thorough testing. 
I did not want to slow down the plan to launch a new version of mana TBH!

PR #437, opened 2025-03-16
Update configure.ac; VERSION set to 1.2.0

gc00:
* Update 'VERSION' file to 1.2.0
 * Check if using Python 3.7 or later
 * Check if dmtcp/configure exists, or else write error msg about need for: git submodule init && git submodule update

NOTE:  We still need to create  aCHANGELOG.md file before the github release.  SEE:
  https://github.com/standard/standard/blob/master/CHANGELOG.md


xuyao0127 (approved):
Let's merge this after other PRs are merged.

gc00:
I'm fine with thatt.  We'll do that almost last.  We also need CHANGELOG.md.

PR #443, opened 2025-03-25
Remove mana_launch/restart.py to mana_launch/restart

xuyao0127:
Remove mana_* scripts and uncomment codes that remove old .mana.rc files


gc00 (approved):
LGTM.

PR #444, opened 2025-03-26
Use getauxval() to check if fsgsbase is available

xuyao0127:


mpi-proxy-split/lower-half/switch-context.cpp
```diff
@@ -45,35 +47,16 @@
 #include <link.h>
 #include <assert.h>
 
-
 #include "switch-context.h"
 #include "lower-half-api.h"
```
gc00, line 52: We need to wrap all of this code in:
```
#ifdef __x86_64__
...
#else
  return 0
#endif
```
We also need to check if MANA still compiles on RISC-V or ARM.  Let's at least test on RISC-V.
We have the assembly instruction "rex.w", which is valid onlly for __x86_64__.  Do we surround the `asm("rex.w")` with `#ifdef __x86_64__`?

Otherwise, we might create bugs for other architectures.  It's because bit 2 of the AT_HWCAP2 word means FSGSBASE only for x86, but not for other architectures..  For other architectures, bit 2 refers to some other hardware feature.  Probably, bit 2 is not yet supported for some other vendor's architecture, but it may be enabled in some future CPU of the other vendor's architecture, and we should not interpret that as calling "rex.w".

[  And we should check DMTCP for these same issues. ]

gc00 (changes requested):
I'm concerned about wrapping this with `#ifdef __x86_64__`.

In addition, I'd like to test if we cal `asm("rex.w ... FSGSBASE")` even on a RISC-V or ARM machine.

gc00 (approved):
LGTM.

PR #446, opened 2025-03-28
Fixed wrong next free addr in mmap_wrapper

xuyao0127:
In the earlier version, the `curr_next_free_addr` in the mmap_wrapper was not updated correctly during restart. As a result, the mmap wrapper mapped new memories at conflicting addresses after restart.


xuyao0127:
I reimplemented the mmap and munmap wrappers in MANA so that it can reuse memories unmapped by the upper-half.

PR #447, opened 2025-03-31
Revert DMTCP PR#1195: bug related to DMTCP init

xuyao0127:
DMTCP's PR#1195 caused a bug when the user program has a constructor and the constructor executes before DMTCP. Before the bug is fixed, MANA reverts the PR in the DMTCP submodule.


gc00 (approved):
LGTM.

PR #448, opened 2025-03-31
mana_launch now detects old ckpt dirs in --ckptdir

xuyao0127:


bin/mana_launch
```diff
@@ -77,6 +78,14 @@ if "--quiet" in dmtcp_flags or "-q" in dmtcp_flags:
   else:
     dmtcp_flags.remove("-q")
 
+# Check if any old checkpoint files exists in the ckptdir
```
gc00, line 81: any old -> an old

gc00 (approved):
Minor inline change.  Otherwise LGTM.

PR #449, opened 2025-04-08
Init flags mana and mpi

rajatpratapbisht:
This PR resolves the inaccurate behavior of `MPI_Initiailized` function call when used with MANA. 

**Cause of Bug**: User MPI application when called `MPI_Initiailized` before calling `MPI_Init` received TRUE flag due to MANA's `MPI_Init` call in lower half.  This inaccuracy in operation was caught when launching GROMACS with MANA.

**Commit Details**:
* `Adds MANA & MPI flags in MPI_Initialized wrapper`: This commit adds new global flags, namely `g_libmpi_is_initialized` and `g_libmana_is_initialized` to keep track of library initialization. 
-- `g_libmpi_is_initialized` is set when User-applciation calls `MPI_Init` marking initialization of MPI_libraries for user.
-- `g_libmana_is_initialized` is set when MPI-plugins are setup in Upper-half before launching User-application.  
When these two flags are set, MPI_Initiliazed call is redirected to lower-half's real MPI-library, otherwise MANA wrapper will respond with flag stating MPI is not Initialized yet. 
  **NOTE:** Choice to call lower-half MPI-library is made purposely to maintain behavioral consistency of MANA as a 'transparent layer' between the user-application and the underlying MPI implementation, ensuring that MPI operations behave as expected when MANA is active
* `New test file for MPI_Initiailized wrapper operation`: This test case is inspired from GROMACS `init` execution which tests for `MPI_Initiailized` before calling `MPI_Init` to avoid reinitialization of MPI environment. This created an issue for MANA since we initialize MPI in lower-half to maintain MPI-agnostic and Network-agnostic behavior.  This test is created to avoid future regression.

### DIFF
--- a/mpi-proxy-split/mpi-wrappers/mpi_wrappers.cpp
+++ b/mpi-proxy-split/mpi-wrappers/mpi_wrappers.cpp
@@ -37,6 +37,7 @@
 #include "uh_wrappers.h"
 
 using namespace dmtcp_mpi;
+bool g_libmpi_is_initialized = false;
 
 static const char collective_p2p_string[] =
    "\n"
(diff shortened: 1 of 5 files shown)


xuyao0127 (approved):
LGTM

PR #450, opened 2025-04-16
Use addr of copied env str in the upper-half stack

xuyao0127:
The previous version of the kernel loader `copy-stack.c` has a bug when copying environment variables to the newly created upper-half stack. We first copied the environment variable strings to the stack, then copied environment variable pointers (`__environ`) to the stack. However, these pointers still point to old copies of the strings in the lower-half stack. Therefore, these pointers became invalid after restart.

This bug makes all environment variables set by `dmtcp_launch` invalid after restart. For example, the "DMTCP_GZIP" controls if the checkpoint image files should be compressed.

In this PR, we fixed the bug by updating environment pointers to correct addresses in the upper-half stack.

### DIFF
--- a/mpi-proxy-split/lower-half/copy-stack.c
+++ b/mpi-proxy-split/lower-half/copy-stack.c
@@ -65,8 +65,10 @@ char *deepCopyStack(int argc, char **argv, char *argc_ptr, char *argv_ptr,
     env_strings_size += strlen(__environ[i]) + 1;
     env_strings_count += 1;
   }
-  int env_ptr_size = (i+1) * sizeof(__environ[0]);
+  int env_ptr_size = (env_strings_count + 1) * sizeof(__environ[0]);
   char *env_ptr_addr = (char *)&__environ[0];
+  char **new_env_ptrs = (char**)malloc(env_ptr_size);
+  memset(new_env_ptrs, 0, env_ptr_size);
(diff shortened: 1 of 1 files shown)


gc00 (approved):
I have a minor change to correct a comment.  After that, LGTM.

mpi-proxy-split/lower-half/copy-stack.c
```diff
@@ -204,10 +212,11 @@ dbg_auxv_ptr_addr = dest_curr_stack;
   // We copied __environ to dest_stack earlier.  So, we're pointing to stack.
   // We use the "padded" variable so that it will be 16-byte aligned.
   dest_curr_stack -= env_ptr_size;
-  memcpy(dest_curr_stack, env_ptr_addr, env_ptr_size);
+  memcpy(dest_curr_stack, new_env_ptrs, env_ptr_size);
```
gc00, line 215: In the line above `memcpy()`, let's add a comment:
```
  // We can't just copy from env_ptr_addr, to dest_curr_stack because
  // env_ptr_addr is an array of pointers to the _old_ stack.  We need
  // new_env_ptrs, an array pointing to the _new_ stack.
```

gc00 (approved):
I have a minor change to correct a comment.  And another comment, also.  After that, LGTM.

PR #453, opened 2025-05-11
Fix for riscv

gc00:
Two commits:
 1.  MANA requires the symtab of ld-linux.so, usually found in the debug version of the file.  We now tell the user if we couldn't find the debug version of the file.
 2. For RISC-V, only, lower-half/lower-half.cpp was fixed to support addresses beyond 2 GB.  The upstream kernel-loader initially assumed static linking within the 2 GB address.  That restriction is now removed int he upstream kernel-loader.


xuyao0127 (approved):
LGTM

PR #454, opened 2025-05-15
Disable checkpointing at MPI_Finalize call

rajatpratapbisht:
When calling `MPI_Finalize`, we disable checkpointing and do not enable it back up. This is done because some MPI applications have special code sections that run for some (but not all) ranks only. 
 
For example, `rank = 0` is generally observed to print out output of Collective Communication calls, such as `MPI_Reduce`. Therefore, since the rest of the ranks  exit before `rank = 0` process, `interval` based checkpointing can create  corrupt checkpoint images, which then give segmentation fault at restart.


xuyao0127 (approved):
LGTM

PR #461, opened 2025-09-04
Update README.md

gc00:
* Update it with a clearer reference to readthdocs.org, and to remove a stale link to mpi-proxy-split/README.md
* Also, add an INSTALL and QUICK-START on this page, for the user's convenience.  It's the same as at readthedocs.orig
* And re-order the citations


xuyao0127 (approved):
LGTM.

PR #466, opened 2026-01-13
Fix segfault in MPI_Type_create_struct

### LINKED ISSUE #465: Segfault when using MPI_Type_create_struct
**Description**: A Segmentation Fault occurs when defining a custom datatype using MPI_Type_create_struct and subsequently using it in communication calls.

**Root Cause**: The issue is caused by incorrect type casting in the record-replay mechanism:

record-replay.h:440
```c
MPI_Datatype *oldtypes = (MPI_Datatype*)(intptr_t)rec->args(3);
```

record-replay.cpp:309
```c
MPI_Datatype *types = (MPI_Datatype*)(intptr_t)rec.args(3);
```

I will address this in an upcoming Merge Request including a new regression test.

dariomnz:
## Fixes #465

Fixes segfault in MPI_Type_create_struct by removing incorrect casts in record-replay.h:440 and record-replay.cpp:309, and adds a missing regression test.

### DIFF
--- a/mpi-proxy-split/record-replay.cpp
+++ b/mpi-proxy-split/record-replay.cpp
@@ -306,7 +306,7 @@ restoreTypeCreateStruct(MpiRecord& rec)
   int count = rec.args(0);
   int *blocklengths = rec.args(1);
   MPI_Aint *displs = rec.args(2);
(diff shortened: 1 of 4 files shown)


xuyao0127 (approved):
LGTM. Thank you for the fix. This will be included in the new release.

PR #473, opened 2026-05-10
Update the mmap wrapper in the lower half

xuyao0127:
This PR improved the mmap wrappers. 

1. mmap() wrapper will always return aligned addresses.
2. Fixed a bug in checking if the mmap flags contains MAP_FIXED.
3. Calculating the mmap arena's base address dynamically. Previous version may have memory conflict if the rlimit (stack size) is too large.


gc00 (approved):
LGTM

PR #474, opened 2026-05-10
Fix Hydra compatibility issue

xuyao0127:
This PR fixed a compatibility issue with Hydra. Hydra uses pipes to communicate with MPI processes. When DMTCP is restoring saved pipes (including pipes created Hydra), it will close them first. Hydra believes the pipe is closed because the MPI process dead. Therefore Hydra will quit. 

Since Hydra's pipes are created when launching the lower-half program, we can temporarily move these pipes to a higher FD and allow DMTCP to restore the pipes it saved. Then before threads resumed, we move these pipes back to the original FDs, overwriting those restored by DMTCP.


gc00 (approved):
LGTM

PR #477, opened 2026-05-29
Updating DMTCP to include the fix for compatibility issue with Hydra

xuyao0127:
The previous PR regarding the Hydra compatibility issue was not a complete solution. Now we updated DMTCP to skip scanning pre-existing sockets if Hydra is used (just like how DMTCP handles SLURM).


gc00 (approved):
LGTM.

PR #478, opened 2026-06-03
Improved P2P drainning algorithm

xuyao0127:
This PR improved the algorithm for checkpointing pending MPI_Recv() functions. 

Previously, we used MPI_Iprobe in a tight loop in the MPI_Recv() wrapper to detect if there's an incoming message. We only calls the real MPI_Recv() function in the lower half if there's a message ready to be received. This has a negative influence on the runtime performance of P2P communications.

In this PR, we removed the loop of MPI_Iprobe() function. Instead, we saved the size, datatype and tag of the to-be-received message until the MPI_Recv() is completed. If the checkpoint request is sent in the middle of a MPI_Recv() function, a matching sender will send a "dummy" message to the pending receiver to unblock it, but still in the wrapper function. After a resume or restart, the wrapper function will call the MPI_Recv function again to receive the real message.

### DIFF
--- a/mpi-proxy-split/mpi-wrappers/mpi_p2p_wrappers.cpp
+++ b/mpi-proxy-split/mpi-wrappers/mpi_p2p_wrappers.cpp
@@ -137,46 +137,112 @@ int PMPI_Rsend(const void* ibuf, int count,
 int PMPI_Recv(void *buf, int count, MPI_Datatype datatype,
              int source, int tag, MPI_Comm comm, MPI_Status *status)
 {
-  int retval;
+  // MANA does not support MPI_THREAD_MULTIPLE.  This wrapper relies on
+  // a single global pending-Recv slot (g_pending_recv) being claimed
+  // by at most one thread at a time.
+  //
+  // Protocol overview (replaces the older MPI_Iprobe polling loop):
+  //
+  //   1. Check the MANA-internal message buffer first.  Messages
+  //      drained from in-flight Sends during a previous checkpoint's
+  //      pre-suspend (via recvMsgIntoInternalBuffer) are served here.
+  //
+  //   2. Publish (source, tag, comm, count, datatype) to g_pending_recv
+  //      with active=true, so that unblockPendingRecvs(), running on
+  //      the DMTCP coordinator thread during a future pre-suspend, can
+  //      identify this rank as blocked and arrange for a matching dummy
+  //      MPI_Send to unblock us.
+  //
+  //   3. Call NEXT_FUNC(Recv) WITHOUT bracketing it in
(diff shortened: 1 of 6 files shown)


mpi-proxy-split/mpi-wrappers/mpi_p2p_wrappers.cpp
```diff
@@ -137,46 +137,112 @@ int PMPI_Rsend(const void* ibuf, int count,
...
+  //      unblockPendingRecvs; discard it, park until mana_state ==
+  //      RUNNING (resume/restart complete), and retry.  Otherwise the
+  //      message is real: increment local_recv_messages, deliver
+  //      status, and return.
+  //
+  // The reason a single post-call read of p2p_dummy_phase is sufficient
+  // is documented at the declaration of p2p_dummy_phase in
+  // p2p_drain_send_recv.h.  Briefly: unblockPendingRecvs only sets
+  // p2p_dummy_phase = true AFTER drainInFlightP2p() exits, which requires
+  // this rank's local_recv_messages to have caught up with sent --
```
gc00: Pls change `sent` to `local_sent_messages` or whatever the code name is for that counter

mpi-proxy-split/p2p_drain_send_recv.cpp
```diff
@@ -314,9 +318,204 @@ consumeMatchingMsgBuffer(void *buf, int count, MPI_Datatype datatype,
...
+    kvdb::get64(db, key, &tag[r]);
+    snprintf(key, sizeof(key), "comm_%d", r);
+    kvdb::get64(db, key, &comm[r]);
+    snprintf(key, sizeof(key), "count_%d", r);
+    kvdb::get64(db, key, &count[r]);
+    snprintf(key, sizeof(key), "dtype_%d", r);
+    kvdb::get64(db, key, &dtype[r]);
+  }
+
+  // Phase C: dispatch dummies.  For each blocked rank j, every member
```
gc00: In the comment, please note in the comment that Phase C is needed _only_ if the pending receive was using MPI_ANY_SOURCE.  Please then write that in this case, one must do a leader election to choose which sender should dispatch the dummy.  And the designated sender will be the lowest rank that is not blocked.
QUESTION:  Is it obvious that there will always be some Sender that is not blocked?  If so, can we have a sentence or two saying why it's obvious.  If it's not obvious, then just make the blanket statement that the algorithm guarantees that there will be a Sender that is not blocked.

gc00 (approved):
1. mpi_p2p_wrappers.cpp : Since we don't support MPI_MULTIPLE_THREADS, MANA should explicitly check if the user is trying to use that, and if so, warn the user.
2. See the inline comments (e.g., add comments if MPI_ANY_SOURCE)

I'm approving it now.

PR #479, opened 2026-06-24
Fixed an address alignment issue in lower-half

xuyao0127:
When the lower-half program load user's program into the memory, if the number of arguments is 8, it will cause a segfault. This is because the $rsp is not 16-bit aligned. This PR fixed the address alignment issue.


gc00 (approved):
LGTM

PR #480, opened 2026-06-25
Fixed deepCopyStack desination address

xuyao0127:
The deepCopyStack function maps a memory region [base, base + statck_size] with MAP_GROWSDOWN flag to set up the new stack. However, it copies data from the existing stack to this new stack from "base" to lower addresses, which means memory region [base - stack_size, base]. This didn't cause a segmentation fault because the MAP_GROWSDOWN makes the kernel extend the memory segment towards lower address, as if it's a stack region.

This PR fixed this issue and renamed some variable names for better readability.


gc00 (approved):
LGTM.  Thanks for fixing this bug that I had created.

PR #484, opened 2026-07-04
Don't skip /dev/zero because upper-half may use it

xuyao0127:
In some versions of CUDA, the checkpoint plugin uses /dev/zero as part of the checkpoint mechanism. Therefore, MANA should not skip these regions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved MPI checkpointing behavior for edge-case memory mappings, reducing the likelihood of incorrectly skipped regions.
  * Refined skip logic for device-backed mappings so `/dev/zero` is no longer treated as a direct “lower-half” indicator, improving preservation of correct memory handling.
* **Chores**
  * Updated the bundled `dmtcp` component to a newer revision.


gc00 (approved):
LGTM

